Issue #1, opened 2020-05-20
Package Design


willtebbutt:
This is intended as a discussion issue where we can hash out an initial design for the package. The goal is to

1. tie down the structure of the package, including how the internals should be designed to lead to make implementing a pleasant user-facing API
2. determine what flavours of approximate inference we want to target first / what is feasible for @sharanry to target over the summer.

None of this is set in stone, so please feel free to chime in with any thoughts you might have on the matter. In particular if you think that I've missed something obvious from the design that could restrict us down the line, now would be a good time to bring it up.

## Background

In an ideal world, the API for GPs with non-Gaussian likelihoods would be "Turing" or "Soss", in the sense that we would just put a GP into a probabilistic programme, and figure out everything from there. This package, however, is not aiming for that level of generality. Rather it is aiming for the tried-and-tested GP + likelihood function API, and providing a robust and well-defined API + collection of approximate inference algorithms to deal with this.

## API

Taking a bottom-up approach to design, my thinking is that the following basic structure should be sufficient for our needs:
```julia
f = GP(m, k)
fx = LatentGP(f, x, ϕ)
log_density(fx, f)
```
where

- `f` is some GP whose inputs are of type `Tx`,
- `x` is some subtype of `AbstractVector{Tx}`,
-  `ϕ` is a function from `AbstractVector{<:Real}` to `Real` that computes the log likelihood a particular sample from `f` at `x`, and
- `log_density(fx, f) := logpdf(fx, f) + ϕ(f)` (it's not clear to me whether this function is ever non-trivial)

This structure encompasses all of the standard things that you'll see in ML, but is a little more general, as the likelihood function isn't restricted to be independent over outputs. To make things convenient for users, we can set up a couple of common cases of `ϕ` such as factorised likelihoods: a type that specifies that `ϕ(f) = sum(n -> ϕ[n](f[n]), eachindex(x))`, and special cases of likelihoods for classification etc (the various things implemented in GPML). I've not figured out _exactly_ what special cases we want here, so we need to put some thought into that.

This interface obviously precludes expressing that the likelihood is a function of entire sample paths from `f` -- see e.g. [1] for an instance of this kind of thing. I can't imagine this being too much of an issue as all of the techniques for actually working with such likelihoods necessarily involve discretising the function, which we can handle. This means that they can still be implemented in an only slightly more ugly manner. If this _does_ turn out to be an actual issue for a number of users, we can always generalise the likelihood a bit.

[...]

PR #3, opened 2020-05-24
Add Basic Structure

sharanry:
#1 

This adds some basic types and functions for `LatentGP`s. Currently, I have added only Gaussian likelihood as a sanity test. I will be adding more likelihoods soon based on your suggestions either in this PR or a separate one.

@willtebbutt


src/latent_gp.jl
```diff
@@ -0,0 +1,30 @@
...
+
+ - `f` is a Gaussian Process.
+ - `x` is an `AbstractVector` of inputs of type `Tx`.
+ - `Φ` is the log likelihood function which maps input of type `Tx` to `Real`.
+    
+"""
+struct LatentGP{T<:GP,Tx}
+    f::T
+    x::AbstractVector{Tx}
+    σ²
```
willtebbutt, line 12: I'm don't think that we need a noise variance parameter for this model, do we? This should be handled by the likelihood now.

edit: this is superseded by my next comment. In general you'll want the noise variance in `fx` to be something close to zero, but typically it won't quite be zero for stability reasons. Either way, we shouldn't have to worry about this too much in this package I don't think, at least not at the API level.

src/latent_gp.jl
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+
+function mean(fx::LatentGP)
+    return AbstractGPs.mean(fx.f, fx.x)
+end
+
+function cov(fx::LatentGP)
+    return AbstractGPs.cov(fx.f, fx.x)
+end
```
willtebbutt, line 30: I don't think that it makes sense to provide these functions in this package. I think we need to discuss this a bit further in #1 to make sure that we're on the same page here.

src/latent_gp.jl
willtebbutt: Also, I think that it makes sense to just keep a single `FiniteGP` object `fx` rather than separately storing `f` and `x`.

sharanry: If we were to keep a single `FiniteGP` object, it would incorporate the noise too. This should solve both the issues.

src/latent_gp.jl
sharanry: If we were to maintain `Distributions.jl` like API, we should probably keep these?

willtebbutt: Ideally we would do this, but it's not generally going to be clear how to compute either the mean or covariance over the output (in the proposal 2 sense). For example if the likelihood is some horrible non-linear thing.

src/latent_gp.jl
```diff
@@ -0,0 +1,32 @@
...
+ - `Φ` is the log likelihood function which maps input of type `Tx` to `Real`.
+    
+"""
+struct LatentGP{T<:AbstractGPs.FiniteGP}
+    fx::T
+    Φ
+    function LatentGP(fx::AbstractGPs.FiniteGP, Φ) where Tx
+        return new{typeof(fx)}(fx, Φ)
+    end    
+end
```
willtebbutt, line 14:
```suggestion
struct LatentGP{Tfx<:AbstractGPs.FiniteGP, TΦ}
    fx::Tfx
    Φ::TΦ
end
```

src/latent_gp.jl
```diff
@@ -0,0 +1,32 @@
...
+
+function LatentGP(f::GP, x::AbstractVector{Tx}, σ², Φ) where Tx
+        return LatentGP(f(x, σ²), Φ)
+end
+
+function LatentGP(f::GP, x::AbstractVector{Tx}, Φ) where Tx
+        return LatentGP(f(x), Φ)
+end
```
willtebbutt, line 22: Do we really need these additional methods? Is it not sufficient to have
```julia
LatentGP(f(x, σ²), Φ)
```
as the unique + correct way to construct a `LatentGP`?

src/likelihoods.jl
```diff
@@ -0,0 +1,10 @@
+
+"""
+    gaussian(fx::LatentGP, y)
+
+Returns gaussian log-likelihood of output `y` w.r.t to the given `LatentGP`.  
+"""
+function gaussian(fx::LatentGP, y)
+    return Distributions.logpdf(MvNormal(mean(fx), cov(fx) + fx.σ² * I), y)
+end
+
```
willtebbutt, line 10: Under the new design, do we need this?

sharanry: Nope. I have more on this for you on slack.

src/latent_gp.jl
sharanry: Thanks. Didn't notice it to be the default constructor. :sweat_smile:

src/latent_gp.jl
sharanry: Nope. Now that you point it out, they seem unnecessary.

Project.toml
```diff
@@ -3,8 +3,19 @@ uuid = "45b822f7-b0f9-4a40-9473-dffa0412f8e7"
...
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
 [compat]
-julia = "1"
+AbstractGPs = "0.1.1"
+Distributions = "0.19, 0.20, 0.21, 0.22, 0.23"
+KernelFunctions = "0.4"
+julia = ">=1.3"
```
willtebbutt, line 18:
```suggestion
julia = "1.3"
```
This means the same thing :)

theogf:
Two general comments :
- I think it would make it generally easier to keep all likelihoods separated (one folder likelihoods and one likelihood per file)
- For likelihoods with a transformation on `f` like Bernoulli or Poisson, it would be nicer to create an additional `link` (like a type) to leave the transformation choice free. (One can then choose between `logit` and `probit` for Bernoulli, `exp`, `square` or `logistic` for Poisson etc

willtebbutt:
> I think it would make it generally easier to keep all likelihoods separated (one folder likelihoods and one likelihood per file)

This seems reasonable to me.

> For likelihoods with a transformation on f like Bernoulli or Poisson, it would be nicer to create an additional link (like a type) to leave the transformation choice free. (One can then choose between logit and probit for Bernoulli, exp, square or logistic for Poisson etc

Agreed that the user should be able to choose the transformation. I don't think that we need to constrain it to be a particular type though -- it's sufficient to let the user just provide a function of their choosing imho.

src/likelihoods.jl
```diff
@@ -0,0 +1,43 @@
...
+```math
+    p(y|f) = Normal(y | f, σ²)
+```
+On calling, this would return a normal distribution with mean `f` and variance σ².
+"""
+struct GaussianLikelihood{T<:Real} <: Likelihood
+    σ²::Vector{T}
+    function GaussianLikelihood(σ²::T) where {T<:Real}
+        new{typeof(σ²)}([σ²])
+    end
```
willtebbutt, line 24:
```suggestion
```

I don't think that we need this.

sharanry: Flux's `params()` seems to fail if the fields are not a vector. I found this kind of a [work around](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/src/basekernels/constant.jl#L51-L56) in `KernelFunctions.jl`

willtebbutt: Oh, sorry, yeah, you're totally right.

theogf: That would be amazing if one could use `Ref`... Do we know if there is any plan on this?

src/trainable.jl
```diff
@@ -0,0 +1,5 @@
+import .Flux.trainable
+
+Flux.trainable(::Likelihood) = ()
```
theogf, line 3: This is not needed, this is already the default behavior

src/latent_gp.jl
```diff
@@ -0,0 +1,29 @@
...
+struct LatentGP{T<:AbstractGPs.FiniteGP}
+    fx::T
+    Φ
+end
+
+function Distributions.rand(rng::AbstractRNG, lgp::LatentGP)
+    f = rand(rng, lgp.fx)
+    y = rand(rng, lgp.Φ(f))
+    return (f=f, y=y)
+end
```
theogf, line 17: I think this would be nice to create `rand` functions also to just get `f` or just get `y`.
Like
```
randf (rng::AbstractRNG, lgp::LatentGP) = rand(rng, lgp.fx)
randy(rng::AbstractRNG, lgp::LatentGP) = rand(rng, lgp).y
```

theogf:
> Agreed that the user should be able to choose the transformation. I don't think that we need to constrain it to be a particular type though -- it's sufficient to let the user just provide a function of their choosing imho.

Well some links can contain hyperparameters, hence the need for a `struct` (a bit in the same way we have `Transform` in `KernelFunctions.jl`.
For example for Poisson one parametrization is `C * logistic(f)` where `C` needs to be optimized.

src/latent_gp.jl
```diff
@@ -0,0 +1,29 @@
+"""
+    LatentGP(f<:GP, x<:AbstractVector, Φ)
+
+ - `fx` is a `FiniteGP`.
+ - `Φ` is the log likelihood function which maps input of type `Tx` to `Real`.
+    
+"""
+struct LatentGP{T<:AbstractGPs.FiniteGP}
+    fx::T
+    Φ
```
devmotion, line 10:
```suggestion
struct LatentGP{T<:AbstractGPs.FiniteGP,S}
    fx::T
    Φ::S
```
Maybe also don't use unicode characters here since it's not just an internal implementation detail but users might want to access the field.

src/latent_gp.jl
willtebbutt: I disagree. What's wrong with
```julia
rand(rng, lgp).f
rand(rng, lgp).y
```
?

theogf: If you are only interested in `f` you allocate `y` for nothing. But I agree that `randy` is not needed

willtebbutt: Sure. If you're worried about performance you can just call
```julia
rand(rng, lgp.fx)
```

willtebbutt: This is not resolved, I really can't see the advantage of having `randf` and `randy`.

willtebbutt:
> Well some links can contain hyperparameters, hence the need for a struct (a bit in the same way we have Transform in KernelFunctions.jl.
For example for Poisson one parametrization is C * logistic(f) where C needs to be optimized.

Sure -- and if the user wants to parametrise it that way that's fine. I'm just saying that we shouldn't require that the linking operation subtypes a particular thing.

src/latent_gp.jl
theogf: I agree we can forget `randy`, but I would still insist that `randf` can be a useful wrapper.

willtebbutt: But why?

theogf:
> Sure -- and if the user wants to parametrise it that way that's fine. I'm just saying that we shouldn't require that the linking operation subtypes a particular thing.

Sounds good. Just wanted to make sure all information was there :grin:

sharanry:
@theogf Do you have any specific "links" in mind? A few basic ones which come to my mind are:
- Scalar
- Logistic
- Softmax
- Compostion


@willtebbutt A random thought. Do you think we can use Bijectors.jl for links? We probably just need a subset of its capabilities.

Also, IMO this should be a separate PR.

src/latent_gp.jl
sharanry: After a second look, I too feel it is better not have `randf` or `randy`. All the components required are exposed and (if required) we could mention this usage in the docstring.

willtebbutt:
> @willtebbutt A random thought. Do you think we can use Bijectors.jl for links? We probably just need a subset of its capabilities.

This sounds reasonable, but I agree that it's something for another PR. In any case, it's also the kind of thing that we can include as a "sensible default", but not something that needs to be enforced.

src/latent_gp.jl
```diff
@@ -0,0 +1,29 @@
+"""
+    LatentGP(f<:GP, x<:AbstractVector, lik)
+
+ - `fx` is a `FiniteGP`.
+ - `lik` is the log likelihood function which maps input of type `Tx` to `Real`.
+
```
devmotion, line 6: The docstring is incorrect (type signature) and the formatting should be fixed.

src/likelihoods/likelihoods.jl
```diff
@@ -0,0 +1,10 @@
+"""
+    Likelihood
+
+An abstract type for likelihoods. Likelihoods are used to model the uncertainities associated
+with the data.
+"""
+abstract type Likelihood end
```
devmotion, line 7: Is the abstract type needed and part of the design? I'm not sure if it's actually useful since no methods are defined for it and it's not part of any type restrictions (and I guess we don't want these restrictions to allow for any kinds of likelihoods).

src/LatentGPs.jl
```diff
@@ -1,5 +1,28 @@
...
+
+export GaussianLikelihood, PoissonLikelihood
+
+
+include("latent_gp.jl")
+include("likelihoods/likelihoods.jl")
+
+function __init__()
+    @require Flux="587475ba-b771-5e3f-ad9e-33799f191a9c" include("trainable.jl")
+end
```
devmotion, line 25: I don't think this is needed anymore (not in KernelFunctions either). The logic was separated into https://github.com/FluxML/Functors.jl. I.e., I guess it would be sufficient to depend on Functors (which should be a light dependency we can take I guess?), and just add
```julia
@functor GaussianLikelihood
```
after the definition of `GaussianLikelihood`.

devmotion: Would be nice to get rid of Requires since it can increase loading times quite a bit, and it would be more generally applicable and not only be restricted to training with Flux.

src/likelihoods/likelihoods.jl
sharanry: It might be useful while defining links which @theogf suggested earlier in [this comment](https://github.com/JuliaGaussianProcesses/LatentGPs.jl/pull/3#issuecomment-634599211).

willtebbutt: How will it help when defining links?

sharanry: This depends on how we design Links. One possible way while making use of `Bijectors.jl` would be
```julia
struct Link
    lik::Likelihood
    transform::Bijector
end
(link::Link)(x) = link.lik(link.transform(x))

#while making a LatentGP
link = Link(PoissonLikelihood(), Exp())
lgp = LatentGP(gp, link)
```
We could maybe come up with something more self-explanatory than "Link".

devmotion: But that's not actually an argument in favour of a `Likelihood` type, you probably just want to have
```julia
struct Link{L,B<:Bijector}
    lik::L
    transform::B
end
```
to allow all kinds of user-defined likelihood functions. `Likelihood` is not used anywhere else in the example, it seems.

sharanry: Yeah you are right. I will remove Likelihood type. Thanks!

willtebbutt:
Sorry for letting this drop off the radar a bit. @sharanry @devmotion @theogf is there anything else in particular that any of you want to get done in this PR? I'm happy for it to be merged I think.

src/LatentGPs.jl
```diff
@@ -1,5 +1,26 @@
...
+using LinearAlgebra
+using Random
+using Functors
+
+import Statistics
+import Distributions
+
+export LatentGP
+
+export logpdf, rand, trainable, randf, randy
```
devmotion, line 15: It seems currently only
```suggestion
export logpdf, rand
```
are defined.

src/likelihoods/gaussian.jl
```diff
@@ -0,0 +1,22 @@
...
+"""
+struct GaussianLikelihood{T<:Real}
+    σ²::T
+end
+
+GaussianLikelihood() = GaussianLikelihood(1e-6)
+
+@functor GaussianLikelihood
+
+(l::GaussianLikelihood)(f::Real) = Normal(f, l.σ²)
```
devmotion, line 20: This should be
```suggestion
(l::GaussianLikelihood)(f::Real) = Normal(f, sqrt(l.σ²))
```

src/likelihoods/gaussian.jl
```diff
@@ -0,0 +1,22 @@
...
+
+(l::GaussianLikelihood)(fs::AbstractVector{<:Real}) = Product([Normal(f, l.σ²) for f in fs])
```
devmotion, line 22: Maybe it might make sense to avoid the iteration and allocation of the intermediate array of distributions and just define
```suggestion
(l::GaussianLikelihood)(fs::AbstractVector{<:Real}) = MvNormal(fs, sqrt(l.σ²))
```

In general I guess one should avoid calling `Product` directly but instead use `product` (same as `Truncated` vs `truncated`) since `product` allows to define (potentially) more efficient dispatches. For instance, here `product` would create just a `MvNormal` distribution.

devmotion:
Maybe one could configure CompatHelper and use a `test/Project.toml` setup right away.

src/likelihoods/gaussian.jl
sharanry: I wasn't aware of the difference between `Product` and `product`. Thank you!

devmotion: Oh I just remembered that it's actually called `product_distribution` (I guess `product` would be a too generic name :smile:).

sharanry:
Does anyone have any objections to merge this?

willtebbutt:
Nope, but I think we'll need to do some refactoring based on #1

sharanry:
> Nope, but I think we'll need to do some refactoring based on #1

Can we do this as a separate PR?

willtebbutt:
Yes, definitely. Please go ahead and merge this :)

Issue #5, opened 2020-08-04
Package name


theogf:
I have the feeling that LikelihoodFunctions is a bit too general for this package. If I understand correctly, this package aims at propose models with Gaussian and non-Gaussian likelihoods. Shouldn't the name be then `GPLikelihoods`, `NonConjugateGPs` or something like this?

yebai:
`NonConjugateGPs` sounds sensible: it reflects well what the package provides, and is self-descriptive I think.

PR #13, opened 2020-09-04
Refactor

sharanry:
This PR 
- removes LatentGPs which has been moved to AbstractGPs.
- Renames the package to NonConjugateGPs as suggested in #5 
- Fixes tests to be compatible with the new design of LatentGPs in AbstractGPs
- Removes dependencies which are not necessary anymore


Project.toml
```diff
@@ -1,5 +1,5 @@
-name = "LatentGPs"
-uuid = "45b822f7-b0f9-4a40-9473-dffa0412f8e7"
+name = "NonConjugateGPs"
+uuid = "6031954c-0455-49d7-b3b9-3e1c99afaf40"
```
devmotion, line 2: I guess we could keep the same UUID?

sharanry: I think the UUIDs are tied to the name deterministically. 

https://discourse.julialang.org/t/problem-with-reassignment-of-uuid-when-a-package-is-registered/11422/2
This is an old post. Not sure how true it is.

sharanry:
If the name "NonConjugateGPs" is not an issue for anyone, I will change the repository name too. 
@devmotion @willtebbutt @theogf @yebai

Project.toml
devmotion: AFAIK this only matters for registered packages (https://github.com/JuliaRegistries/General#how-do-i-rename-an-existing-registered-package).

devmotion: And this package is not registered yet?

willtebbutt:
What functionality are we actually planning to provide in this package? Just the likelihoods? Or approximate inference as well? 

I'm also somewhat dubious about calling this "NonConjugateGPs" because I don't know what that means. i.e. what _is_ a non-conjugate Gaussian process?

Project.toml
sharanry: I didn't realize that the package hasn't been registered yet. 😅 In that case, the change in the UUID shouldn't matter? Am I missing something?

devmotion: True :smile:

theogf:
Fair enough I am so usee to the term that I forgot it is a very big shortcut. A non-conjugate GP is for me a Gaussian Process on which you do not apply a Gaussian likelihood.

sharanry:
If I understand correctly, all we need for non-Gaussian GPs is the likelihoods defined in this package along with the GP and LatentGPs given by AbstractGPs. 

Am I right in saying `LatentGP` could be a non-Gaussian GP? Or is there some other component I am missing?

If this is true, the likelihoods defined in this package are quite generic and are not tied GPs in the way they are defined. `LikelihoodFunctions.jl` is probably an fair name to have. 

However, our tests are from the context of GPs. Should we change this to be more generic? If we do, I am not sure what would be the right place add such GP related tests.

yebai:
Maybe call this “GPML2” since it is a user-facing libraries? The digit in the name is to satisfy Julia package conventions.

yebai:
I am also open to recycling the name “GPMLJ”, which is a good name, I think.

theogf:
I quite dislike GPML(2/J) because it is very uninformative and anyway packages guidelines recommend to avoid acronyms.
If I understood correctly, we only want to have the likelihoods implemented in here but not the inference for it. IMHO the name `LikelihoodFunctions` is just super generic and maybe too much for our aim. How about simply `GPLikelihoods.jl`?

willtebbutt:
> If I understood correctly, we only want to have the likelihoods implemented in here but not the inference for it. IMHO the name LikelihoodFunctions is just super generic and maybe too much for our aim. How about simply GPLikelihoods.jl?

`GPLikelihoods.jl` is something that I could get on board with -- it's clear that we're talking about likelihoods for GPs.

yebai:
My feeling is that these likelihood functions will be highly intertwined with inference methods. So it might be hard to have something as clean as KernelFunctions. So the scope of this package will likely become likelihood functions and a large part of approximate inference. If that’s the case, then a slightly more generic name that includes approximate inference might make sense. 

I don't see why acronyms are bad, for example Stats, MCMC, GP are used ubiquitously.

willtebbutt:
> I don't see why acronyms are bad, for example Stats, MCMC, GP are used ubiquitously.

They're not inherently bad -- to my mind descriptive names are just preferable because they're descriptive. You have to explain acronyms to people, and they can be easy to forget if they're not obvious.

For example, I could probably get on board with calling something `GPML2` if it's doing something similar to GPML.

I could not get on board with calling something `GPMLJ.jl`, because the `J` is repeating information already present in the `.jl`. The fact that `MLJ.jl` does it does not mean that we should do it. Their naming decision was bad.

theogf:
@willtebbutt Given what @yebai is saying it is not clear to me what is the aim for this package:
I understand that we want to provide functions to provide likelihoods compatible with the AbstractGPs framework and more generally with the Turing framework.
What I don't understand is: do we also want to provide inference for it? In my opinion opening Pandora's box as there are sooooo many ways to do inference for GPs...
I would then rather see an additional package namely "NonConjugateGPs" providing special inference methods as well as notebook examples on how to do this more "classically", i.e. use a Turing model and sample from it, or use AdvancedVI.jl etc...

willtebbutt:
Yeah. I was under the impression that this was going to be a place where we specify a bunch of likelihoods that are to be used with `LatentGPs` in `AbstractGPs`. The reason for keeping them separate being a concern about the dependencies that this package is likely to introduce.

I was under the impression that a separate "GPML"-like package was going to be the thing that ties this together for users.

theogf:
Ok so let's make this democratic :) 
 - :+1: NonConjugateGPs
- :heart: GPLikelihoods
- :laughing: LikelihoodFunctions
- :tada: GPML2
- :rocket: GPMLJ

src/GPLikelihoods.jl
```diff
@@ -1,24 +1,15 @@
-module LatentGPs
+module GPLikelihoods
 
 using Distributions
 using KernelFunctions
```
theogf, line 4: I don't think `KernelFunctions` is used anywhere either

test/Project.toml
```diff
@@ -3,12 +3,12 @@ AbstractGPs = "99985d1d-32ba-4be9-9821-2ec096f28918"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
```
theogf, line 5: cf previous comment

yebai:
I am happy with the `GPLikelihoods` name. 

Still, I think it's prudent to develop the `likelihood` interface for a couple of weeks before embarking registration. See the reasoning below. 

> The reason for keeping them separate being a concern about the dependencies that this package is likely to introduce.

My feeling is that the need for a dedicated likelihood package is not big since a lot of the functionality exist in `Distributions`. The extra bits needed for GP are likely some user-friendly interfaces.  Therefore, the need for a separate package might fade if the `likelihood` types and methods can be sufficiently abstract. In that case, we can consider making an effort to avoid explicit dependence on heavy packages like `Distributions`. Then we can keep the generic `likelihood` types and methods inside the `AbstractGPs` package. 

This is potentially better in the long run because it allows GP inference developers to extend this likelihood interface by importing the `AbstractGPs` package. For example, each inference method can be in its own package, providing inference specific extensions of likelihood functions. On top of that, we can have a user-facing library (similar to the `GPML` Matlab library, or GPFlow) that depends on `AbstractGPs`, `KernelFunctions` and various inference packages.

test/runtests.jl
```diff
@@ -1,15 +1,12 @@
-using LatentGPs
+using GPLikelihoods
 using KernelFunctions
```
theogf, line 2: same here

theogf (commented):
Looks good but I think you can also remove KernelFunctions from the dependencies

theogf (approved):

PR #14, opened 2020-09-06
Adding a Link object

theogf:
We need to support different type of mapping when a constraint is applied on the GP, e.g. for Poisson the given standard was `exp(f)` but there are more existing options : `f^2`, or `s * logistic(f)`

The convention is to call the connection between the parameter(s) `m` needed for the likelihood and the input(s) `f` the **inverse link**. Cause traditionally the **link** would be the function to go from `m` to `f`. See for example : http://web.pdx.edu/~newsomj/mvclass/ho_link.pdf

This PR aims at introducing this with  `AbstractLink` type.
Some likelihoods **need** an inverse link, for example `BernoulliLikelihood` or `PoissonLikelihood`. So it makes sense to have these likelihoods contain the inverse link `invlink` and parametrize it. This allows furthermore to differentiate different overloads later and for people to use their own transformations. For example my augmentation stuff only works for the `LogitLink` but not the `ProbitLink`.

For a small collection of `AbstractLink`s I defined just `apply(link, x)` (which is called by `(link::LinkType)(x)`. and also `Base.inv` which returns the existing inverse link if it exists (similarly to Bijectors).

I already did a basic implementation but I realised a lot of it is overlapping with [`Bijectors.jl`](https://github.com/TuringLang/Bijectors.jl) and we could consider using them. However this is also a heavier dependency ? I think `NNlib` is pretty heavy right?


devmotion:
An alternative would be to not use different link functions but define additional likelihood functions (i.e., one for Poisson with the `exp` link, one with `x^2` etc.).

theogf:
I thought that it would be cooler to be able to input really any kind of link! One could even think about passing a constrained NN (paper idea here :laughing: )

src/likelihoods/gaussian.jl
```diff
@@ -10,13 +10,15 @@ uncertainity associated with the data follows a Gaussian distribution.
 On calling, this would return a normal distribution with mean `f` and variance σ².
 """
 struct GaussianLikelihood{T<:Real}
-    σ²::T
+    σ²::Vector{T}
```
sharanry, line 13: Do we need things to be in a vector form if we are using `Functors`?

sharanry: Also, shouldn't we have a link here too? We should be able to use Links regardless of whether the parameters of a distribution are constrained.

sharanry: `IdentityLink` could be the default.

src/likelihoods/link.jl
```diff
@@ -0,0 +1,21 @@
+abstract type AbstractLink end
+
+(f::AbstractLink)(x) = apply(f, x)
+
+struct Link{F} <: AbstractLink
+    f::F
+end
+
```
sharanry, line 8: As mentioned before I think it is good to have a identity link for the sake of completeness.

sharanry (commented):
Is it possible to keep links separated from the likelihoods? That is, not have it part of every likelihood? We could think of this as an optional step between the GP outputs and likelihood function.

We could also define `LinkedLikelihood` similar to transforms in `KernelFunctions.jl` such that it is just another likelihood.
```julia
struct LinkedLikelihood{T}
link::Link
lik::T
end

(ll::LinkedLikelihood)(f::Real) = ll.lik(ll.link(f))
(ll::LinkedLikelihood)(fs::AbstractVector{<:Real}) = ll.lik(ll.link.(fs))
```
or something like
```julia
struct LinkedLikelihood{T1,T2}
link::T1
lik::T2
end

(ll::LinkedLikelihood)(f::Real) = ll.lik(ll.link(f))
(ll::LinkedLikelihood)(fs::AbstractVector{<:Real}) = ll.lik(ll.link.(fs))
```
I am supportive of the second approach as both links and likelihoods are essentially functions. And the second approach provides the maximum amount of flexibility.

src/likelihoods/gaussian.jl
theogf: Here a link is not needed because `f` is just used as the mean of the Gaussian and is already in the right domain

theogf:
> Is it possible to keep links separated from the likelihoods? That is, not have it part of every likelihood? We could think of this as an optional step between the GP outputs and likelihood function.
> 
> We could also define `LinkedLikelihood` similar to transforms in `KernelFunctions.jl` such that it is just another likelihood.

I don't think it is possible to separate links and likelihoods. Unlike in `KernelFunctions`, for some `Likelihoods` links/transformations are **necessary**. I like the idea of `LinkedLikelihood` to have link for `Likelihood` which don't need a `Link` but for which it could be interesting to have one.

sharanry:
> Unlike in `KernelFunctions`, for some `Likelihoods` links/transformations are **necessary**.

Okay that makes sense. 🙂 But we should always have a sensible default link in such cases.

theogf:
Actually one change that could be need is the naming. Should it be `InvLink` or `Link`? For example in GPFlow they use `invlink` : https://github.com/GPflow/GPflow/blob/5a945d67b37120610880c3323224a4e86404ae1d/gpflow/likelihoods/scalar_discrete.py#L16
Additionally here: https://en.wikipedia.org/wiki/Generalized_linear_model#Link_function, the link is defined as `logit` and not `logistic`. Never really understood why btw...

src/likelihoods/gaussian.jl
theogf: The Vector form is to insure compatibility with Zygote

sharanry: > Here a link is not needed because f is just used as the mean of the Gaussian and is already in the right domain

Yes but I just wanted to give that flexibility. I think `LinkedLikelihood` will serve that purpose.

theogf:
As an alternative (I feel I am rewriting existing code) we could use `Bijectors.jl` To create the links?
I think for the most important ones, there exist a bijector and its inverse for it.

devmotion:
IMO we shouldn't use Bijectors.jl, mainly because it's a very heavy dependency and, more importantly, since my impression is that its design is very much focused on Turing and AdvancedHMC. In particular, all inputs and outputs have to be arrays and have to be of the same dimension and size. IMO this is quite limiting in more general applications where you want to map a submanifold of lower dimension (e.g. matrices with a special structure or the simplex - there are some open issues but I don't think the design will be changed in the near future since it would be very breaking for Turing).

However, I also think it would be good to use existing functionality in other packages, if possible. Maybe TransformVariables could be helpful? IIRC it supports more flexible mappings and input and output types and would be a lighter dependency.

theogf:
I had a look at TransformVariables and it looks great indeed, only I think we would need to use the low level API the `as` API is too limiting. Funnily enough they have a `transform` function 😂. Sounds like the change for KernelFunctions.jl comes right in time!

src/GPLikelihoods.jl
```diff
@@ -4,7 +4,7 @@ using Distributions
 using AbstractGPs
 using Random
 using Functors
-using StatsFuns: logistic, softmax
+using StatsFuns: logistic
```
devmotion, line 7: We could think about using LogExpFunctions instead: https://github.com/JuliaStats/StatsFuns.jl/pull/108

st--:
> Actually one change that could be need is the naming. Should it be `InvLink` or `Link`? For example in GPFlow they use `invlink` : https://github.com/GPflow/GPflow/blob/5a945d67b37120610880c3323224a4e86404ae1d/gpflow/likelihoods/scalar_discrete.py#L16
> Additionally here: https://en.wikipedia.org/wiki/Generalized_linear_model#Link_function, the link is defined as `logit` and not `logistic`. Never really understood why btw...

Depends on which way around you define the bijector. In the statistics community, the link function is well defined to be the object that satisfies link(y) = f, where y is your observations/likelihood parameters, and f is your latent (generally assumed linear) model. So y = invlink(f). In implementing likelihoods for our GP models we generally care about the f -> y direction only, so we define the invlink explicitly.

theogf:
@st-- Thanks for the explanation. I was indeed not familiar with the statistics nomenclature. The names have been changed accordingly already. One can switch from one representation to another (when possible) with `inv`

src/likelihoods/categorical.jl
```diff
@@ -9,8 +9,12 @@ uncertainity associated with the data follows a Categorical distribution.
 On calling, this would return a Categorical distribution with `f_i` 
 probability of `i` category.
 """
-struct CategoricalLikelihood end
+struct CategoricalLikelihood{Tl<:AbtractLink}
+    invlink::Tl
+end
 
-(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(softmax(vcat(f, 0)))
+CategoricalLikelihood(invlink::T=SoftMaxLink()) where{T<:AbstractLink} = CategoricalLikelihood{T}(invlink)
```
willtebbutt, line 16:
```suggestion
CategoricalLikelihood() = CategoricalLikelihood(SoftMaxLink())
```
You get the single argument version for free frm the default constructor.

willtebbutt (commented):
Just done a quick pass.

theogf:
@st-- @willtebbutt 
Ready for another pass

src/likelihoods/categorical.jl
```diff
@@ -1,16 +1,20 @@
...
+    invlink::Tl
+end
 
-(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(softmax(vcat(f, 0)))
+CategoricalLikelihood() = CategoricalLikelihood(SoftMaxLink())
 
-(l::CategoricalLikelihood)(fs::AbstractVector) = Product(Categorical.(softmax.(vcat.(fs, 0))))
+(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(l.invlink(vcat(f, 0)))
+
+(l::CategoricalLikelihood)(fs::AbstractVector) = Product(Categorical.(l.invlink.(vcat.(fs, 0))))
```
willtebbutt, line 20: Line length

src/likelihoods/categorical.jl
```diff
@@ -1,16 +1,20 @@
...
 Categorical likelihood is to be used if we assume that the 
 uncertainity associated with the data follows a Categorical distribution.
 ```math
-    p(y|f_1, f_2, \\dots, f_{n-1}) = Categorical(y | softmax(f_1, f_2, \\dots, f_{n-1}, 0))
+    p(y|f_1, f_2, \\dots, f_{n-1}) = Categorical(y | l(f_1, f_2, \\dots, f_{n-1}, 0))
 ```
-On calling, this would return a Categorical distribution with `f_i` 
-probability of `i` category.
+On calling, this would return a Categorical distribution where probabilities are 
+given by `l(f_1, f_2, ..., f_{n-1}, 0)`, a vector of probabilities, which sums to 1
```
willtebbutt, line 10: I wonder whether this should be a little bit more literal, and make it clear that you need to provide a vector, not a tuple, as input?

src/likelihoods/exponential.jl
```diff
@@ -1,14 +1,18 @@
...
+    invlink::Tl
+end
 
-(l::ExponentialLikelihood)(f::Real) = Exponential(exp(f))
+ExponentialLikelihood() = ExponentialLikelihood(ExpLink())
 
-(l::ExponentialLikelihood)(fs::AbstractVector{<:Real}) = Product(Exponential.(exp.(fs)))
+(l::ExponentialLikelihood)(f::Real) = Exponential(l.invlink(f))
+
+(l::ExponentialLikelihood)(fs::AbstractVector{<:Real}) = Product(Exponential.(l.invlink.(fs)))
```
willtebbutt, line 18: Line length

src/likelihoods/gamma.jl
```diff
@@ -1,21 +1,24 @@
 """
-    GammaLikelihood(α)
+    GammaLikelihood(α::Real=1.0, l::AbstractLink=ExpLink())
 
 Gamma likelihood with fixed shape `α`.
 
 ```math
-    p(y|f) = Gamma(y | α, θ=exp(f))
+    p(y|f) = Gamma(y | α, θ=l(f))
```
willtebbutt, line 7: I'm not sure that having the \theta character here doesn't create more confusion than clarity. Below is states that `l(f)` is the scale, which I would imagine is sufficient?

src/likelihoods/gaussian.jl
```diff
@@ -10,31 +10,38 @@ uncertainity associated with the data follows a Gaussian distribution.
...
 """
-struct HeteroscedasticGaussianLikelihood end
+struct HeteroscedasticGaussianLikelihood{Tl<:AbstractLink}
+    invlink::Tl
+end
+
+HeteroscedasticGaussianLikelihood() = HeteroscedasticGaussianLikelihood(ExpLink())
 
-(::HeteroscedasticGaussianLikelihood)(f::AbstractVector{<:Real}) = Normal(f[1], exp(f[2]))
+(l::HeteroscedasticGaussianLikelihood)(f::AbstractVector{<:Real}) = Normal(f[1], l.invlink(f[2]))
```
willtebbutt, line 45: Line length

src/likelihoods/gaussian.jl
```diff
@@ -10,31 +10,38 @@ uncertainity associated with the data follows a Gaussian distribution.
...
 
-(::HeteroscedasticGaussianLikelihood)(fs::AbstractVector) = MvNormal(first.(fs), exp.(last.(fs)))
+(l::HeteroscedasticGaussianLikelihood)(fs::AbstractVector) = MvNormal(first.(fs), l.invlink.(last.(fs)))
```
willtebbutt, line 47: Line length

src/likelihoods/gaussian.jl
```diff
@@ -10,31 +10,38 @@ uncertainity associated with the data follows a Gaussian distribution.
...
 Heteroscedastic Gaussian likelihood. 
 This is a Gaussian likelihood whose mean and the log of whose variance are functions of the
 latent process.
 
 ```math
-    p(y|[f, g]) = Normal(y | f, exp(g))
+    p(y|[f, g]) = Normal(y | f, l(g))
 ```
-On calling, this would return a normal distribution with mean `f` and variance `exp(g)`.
+On calling, this would return a normal distribution with mean `f` and variance `l(g)`.
```
willtebbutt, line 36: Should this be variance or standard deviation?

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+    AbstractLink
+
+Abstract type defining maps from R^n -> X.
+They can be applied by calling `link(x)`
+
+A series of definitions are given in http://web.pdx.edu/~newsomj/mvclass/ho_link.pdf
+"""
+abstract type AbstractLink end
+
+(f::AbstractLink)(x) = apply(f, x)
```
willtebbutt, line 11: Is there a particular reason for defining `apply`, rather than just implementing `(f::ConcreteLink)(x)` directly?

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+struct LogLink <: AbstractLink end
+
+apply(::LogLink, x) = log(x)
+
+Base.inv(::LogLink) = ExpLink()
+
+"""
+    ExpLink()
+
+`exp` link. f:ℝ->ℝ⁺. Its inverse is the [`ExpLink`](@ref)
```
willtebbutt, line 45:
```suggestion
`exp` link. f:ℝ->ℝ⁺. Its inverse is the [`LogLink`](@ref)
```
?

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+
+apply(::InvLink, x) = inv(x)
+
+Base.inv(::InvLink) = InvLink()
+
+"""
+    SqrtLink()
+
+`sqrt` link. f:ℝ⁺∪{0}->ℝ⁺∪{0}.
+Its inverse is the [`SquareLink`](@ref)
```
willtebbutt, line 68:
```suggestion
Its inverse is the [`SquareLink`](@ref).
```

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+
+apply(::SqrtLink, x) = sqrt(x)
+
+Base.inv(::SqrtLink) = SquareLink()
+
+"""
+    SquareLink()
+
+`^2` link. f:ℝ->ℝ⁺∪{0}.
+Its inverse is the [`SqrtLink`](@ref)
```
willtebbutt, line 80:
```suggestion
Its inverse is the [`SqrtLink`](@ref).
```

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+struct SquareLink <: AbstractLink end
+
+apply(::SquareLink, x) = x^2
+
+Base.inv(::SquareLink) = SqrtLink()
+
+"""
+    LogitLink()
+
+Determines the `logit` link, i.e `log(x/(1-x))`, f:[0,1]->ℝ
```
willtebbutt, line 91:
```suggestion
`log(x/(1-x))` link. f:[0,1]->ℝ.
```
For consistency.

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+struct LogitLink <: AbstractLink end
+
+apply(::LogitLink, x) = logit(x)
+
+Base.inv(::LogitLink) = LogisticLink()
+
+"""
+    LogisticLink()
+
+Determines the `logistic` link, i.e `exp(x)/(1+exp(-x))`. f:ℝ->[0,1]
```
willtebbutt, line 103:
```suggestion
`exp(x)/(1+exp(-x))` link. f:ℝ->[0,1].
```
For consistency.

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+struct ProbitLink <: AbstractLink end
+
+apply(::ProbitLink, x) = norminvcdf(x)
+
+Base.inv(::ProbitLink) = NormalCDFLink()
+
+"""
+    NormalCDFLink()
+
+Determines the `cdf` of a normal, i.e ∫₋∞ˣ p(x')dx'. f:ℝ->[0,1]
```
willtebbutt, line 121:
```suggestion
Determines the `cdf` of a normal, i.e ∫₋∞ˣ p(x')dx'. f:ℝ->[0,1].
```

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+apply(::NormalCDFLink, x) = normcdf(x)
+
+Base.inv(::NormalCDFLink) = ProbitLink()
+
+apply(::ChainLink{<:Tuple{LogLink, NormalCDFLink}}, x) = normlogcdf(x) # Specialisation for log + normal cdf
+
+"""
+    SoftMaxLink()
+
+Determines the `softmax` link, i.e `exp(f)/(1+exp(-x))`. f:ℝⁿ->Sⁿ. Where Sⁿ is an n-simplex : TODO to check
```
willtebbutt, line 135: Which aspect of this remains TODO?

src/links.jl
```diff
@@ -0,0 +1,139 @@
...
+struct Link{F} <: AbstractLink
+    f::F
+end
+
+apply(l::Link, x) = l.f(x)
+
+"""
+    LogLink()
+
+`log` link: f:ℝ⁺->ℝ . Its inverse is the [`ExpLink`](@ref)
```
willtebbutt, line 34:
```suggestion
`log` link. f:ℝ⁺->ℝ . Its inverse is the [`ExpLink`](@ref).
```

test/links.jl
```diff
@@ -0,0 +1,54 @@
...
+
+    # Generic link
+    f = sin
+    l = Link(f)
+    @test l(x) == f(x)
+
+    # Log
+    l = LogLink()
+    @test l(x) == log(x)
+    @test inv(l) == ExpLink()
```
willtebbutt, line 13: Could you add another test to each of these (that define inverse links) to ensure that if you call the inverse link twice, you get the same kind of link back?

willtebbutt (commented):
Broadly looks really nice -- a number of style things / comments on docstrings, and a couple of small design things.

src/links.jl
theogf: Not that this is complicated here, but isn't it complicated to give ChainRules for functor functions?

willtebbutt: It shouldn't be. I mean, all `rrule`s are of the form
```julia
rrule(config, f, args...)
```
or
```julia
rrule(f, args...)
```
and I think the testing infrastructure now handles this properly (Lyndon and Miha at Invenia have been working full time on ChainRules for the last couple of months -- they've made loads of progress and it's really very exciting).

theogf: Ok I will get read of `apply`

willtebbutt (approved):
I'm happy with this now.

PR #15, opened 2020-09-07
Add Bernoulli likelihood

sharanry:
Adds Bernoulli likelihood which would enable us to easily model binary classification tasks.

### DIFF
--- a/src/likelihoods/bernoulli.jl
+++ b/src/likelihoods/bernoulli.jl
@@ -11,8 +11,6 @@ On calling, this would return a Bernoulli distribution with `f` probability of `
 """
 struct BernoulliLikelihood end
 
-@functor BernoulliLikelihood
-
 (l::BernoulliLikelihood)(f::Real) = Bernoulli(logistic(f))
 
 (l::BernoulliLikelihood)(fs::AbstractVector{<:Real}) = Product(Bernoulli.(logistic.(fs)))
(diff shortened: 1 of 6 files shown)


src/likelihoods/bernoulli.jl
```diff
@@ -0,0 +1,20 @@
...
+```math
+    p(y|f) = Bernoulli(y | f)
+```
+On calling, this would return a Bernoulli distribution with `f` probability of `true`.
+"""
+struct BernoulliLikelihood end
+
+@functor BernoulliLikelihood
+
+_logistic(logit::Real) = 1/(1 + exp(-logit))
```
theogf, line 16: You can do like in #14 and use `logistic` from `StatsFuns`

sharanry:
Is it okay to merge this?

src/likelihoods/bernoulli.jl
```diff
@@ -0,0 +1,18 @@
...
+uncertainity associated with the data follows a Bernoulli distribution.
+
+```math
+    p(y|f) = Bernoulli(y | f)
+```
+On calling, this would return a Bernoulli distribution with `f` probability of `true`.
+"""
+struct BernoulliLikelihood end
+
+@functor BernoulliLikelihood
```
devmotion, line 14:
```suggestion
```

devmotion:
Same here, I'm not sure if @willtebbutt's intention was that https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/issues/18 should be addressed first.

willtebbutt:
Same here. Feel free to merge and refactor one test utils are sorted.

PR #16, opened 2020-09-17
Add Categorical Likelihood

sharanry:
Adds categorical likelihood. This can be used along with `MOInput`s and `MOKernel`s to model multi-class classification.

### DIFF
--- a/src/likelihoods/categorical.jl
+++ b/src/likelihoods/categorical.jl
@@ -1,0 +1,18 @@
+
+"""
+    CategoricalLikelihood
+
+Categorical likelihood is to be used if we assume that the 
+uncertainity associated with the data follows a Categorical distribution.
+```math
+    p(y|f_1, f_2, \\dots, f_n) = Categorical(y | f_1, f_2, \\dots, f_n)
+```
+On calling, this would return a Categorical distribution with `f` probability of `true`.
+"""
+struct CategoricalLikelihood end
+
+@functor CategoricalLikelihood
+
+(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(softmax(f))
+
+(l::CategoricalLikelihood)(fs::AbstractVector) = Product(Categorical.(softmax.(fs)))
--- a/src/likelihoods/categorical.jl
+++ b/src/likelihoods/categorical.jl
@@ -7,7 +7,8 @@ uncertainity associated with the data follows a Categorical distribution.
 ```math
     p(y|f_1, f_2, \\dots, f_n) = Categorical(y | f_1, f_2, \\dots, f_n)
 ```
-On calling, this would return a Categorical distribution with `f` probability of `true`.
+On calling, this would return a Categorical distribution with `f_i` 
+probability of `i` category.
 """
 struct CategoricalLikelihood end
 
--- a/src/likelihoods/categorical.jl
+++ b/src/likelihoods/categorical.jl
@@ -11,6 +11,6 @@ probability of `i` category.
 """
 struct CategoricalLikelihood end
 
-(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(softmax(append!(f, 0)))
+(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(softmax(vcat(f, 0)))
 
-(l::CategoricalLikelihood)(fs::AbstractVector) = Product(Categorical.(softmax.(append!.(fs, 0))))
+(l::CategoricalLikelihood)(fs::AbstractVector) = Product(Categorical.(softmax.(vcat.(fs, 0))))
--- a/src/likelihoods/categorical.jl
+++ b/src/likelihoods/categorical.jl
@@ -1,18 +1,15 @@
-
 """
     CategoricalLikelihood
 
 Categorical likelihood is to be used if we assume that the 
 uncertainity associated with the data follows a Categorical distribution.
 ```math
-    p(y|f_1, f_2, \\dots, f_n) = Categorical(y | f_1, f_2, \\dots, f_n)
+    p(y|f_1, f_2, \\dots, f_n) = Categorical(y | softmax(f_1, f_2, \\dots, f_n))
 ```
 On calling, this would return a Categorical distribution with `f_i` 
 probability of `i` category.
 """
 struct CategoricalLikelihood end
-
-@functor CategoricalLikelihood
 
 (l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(softmax(f))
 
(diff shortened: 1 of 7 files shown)


src/likelihoods/categorical.jl
```diff
@@ -0,0 +1,19 @@
+
```
devmotion, line 1:
```suggestion
```

src/likelihoods/categorical.jl
```diff
@@ -0,0 +1,19 @@
...
+uncertainity associated with the data follows a Categorical distribution.
+```math
+    p(y|f_1, f_2, \\dots, f_n) = Categorical(y | f_1, f_2, \\dots, f_n)
+```
+On calling, this would return a Categorical distribution with `f_i` 
+probability of `i` category.
+"""
+struct CategoricalLikelihood end
+
+@functor CategoricalLikelihood
```
devmotion, line 15: There are no parameters here, so no need to use `@functor`:
```suggestion
```

devmotion: If you don't define it, `functor` just falls back to https://github.com/FluxML/Functors.jl/blob/fa1a212702fe34f06b3cb18ca2fd260c9c6dcd4c/src/functor.jl#L1-L2

test/likelihoods/categorical.jl
```diff
@@ -0,0 +1,21 @@
...
+    lik = CategoricalLikelihood()
+    lgp = LatentGP(gp, lik, 1e-5)
+    lfgp = lgp(X)
+
+    Y = rand(rng, lfgp.fx)
+    
+    y = [Y[[i + j*N for j in 0:(OUT_DIM - 1)]] for i in 1:N]
+    # Replace with mo_inverse_transform once it is merged
+
+    @test typeof(lik(y)) <: Distribution
```
devmotion, line 18: Only a personal preference (so feel free to not change it)
```suggestion
    @test lik(y) isa Distribution
```

src/likelihoods/categorical.jl
```diff
@@ -0,0 +1,19 @@
+
+"""
+    CategoricalLikelihood
+
+Categorical likelihood is to be used if we assume that the 
+uncertainity associated with the data follows a Categorical distribution.
+```math
+    p(y|f_1, f_2, \\dots, f_n) = Categorical(y | f_1, f_2, \\dots, f_n)
```
theogf, line 8:
```suggestion
    p(y|f_1, f_2, \\dots, f_n) = Categorical(y | softmax(f_1, f_2, \\dots, f_n))
```

src/likelihoods/categorical.jl
sharanry: Wasn't aware of this. Thank you!

test/likelihoods/categorical.jl
sharanry: Even I prefer `isa` 😅 but all the tests till now seem to be using `typeof`. I will change them.

src/likelihoods/categorical.jl
```diff
@@ -0,0 +1,16 @@
+"""
+    CategoricalLikelihood
+
+Categorical likelihood is to be used if we assume that the 
+uncertainity associated with the data follows a Categorical distribution.
+```math
+    p(y|f_1, f_2, \\dots, f_n) = Categorical(y | softmax(f_1, f_2, \\dots, f_n))
```
devmotion, line 7: I just noticed that this will create unidentifiability issues (the resulting distribution is invariant under additions of constant vectors). These could be resolved by e.g. defining instead
```suggestion
    p(y | f_2, f_3, \\dots, f_n) = Categorical(y | softmax(1, f_2, \\dots, f_n))
```
(basically `f_1` is fixed to 1).

sharanry: Isn't this a bit non-intuitive? Is there no other way to avoid in-variances?

devmotion: That's the common approach, e.g., in [multinomial logistic regression](https://en.wikipedia.org/wiki/Multinomial_logistic_regression) (on Wikipedia `f_n` is set to 1 instead of `f_1`).

theogf: It's in the same way that for classifying a binary case you only need one GP, not two.
I would however be in favor of keeping the convention for the wiki page : 
```suggestion
    p(y|f_1, f_2, \\dots, f_n) = Categorical(y | softmax(f_1, f_2, \\dots, f_{n-1}, 1))
```

sharanry: I was just checking other packages to see what they have done.

GPFlow doesn't seem to have an equivalent implementation for Categorical likelihood. In GPyTorch, I think `SoftmaxLikelihood` is the equivalent. https://docs.gpytorch.ai/en/latest/_modules/gpytorch/likelihoods/softmax_likelihood.html#SoftmaxLikelihood

Weirdly they don't seem to be worried about invariances but they optionally support "mixing weights".

sharanry: My question was: Do we want to support something like to "mixing weights"?

theogf: It's a bit weird from this line 
`mixed_fs = function_samples @ self.mixing_weights.t()`
It sounds like they just make a linear combinations of the output (already what we can do with the kernel basically), I would not consider it.

devmotion: BTW it should of course be `0` in the softmax function (I'm not sure why I wrote 1):
```suggestion
    p(y|f_1, f_2, \\dots, f_{n-1}) = Categorical(y | softmax(f_1, f_2, \\dots, f_{n-1}, 0))
```

src/likelihoods/categorical.jl
```diff
@@ -0,0 +1,16 @@
...
+uncertainity associated with the data follows a Categorical distribution.
+```math
+    p(y|f_1, f_2, \\dots, f_{n-1}) = Categorical(y | softmax(f_1, f_2, \\dots, f_{n-1}, 0))
+```
+On calling, this would return a Categorical distribution with `f_i` 
+probability of `i` category.
+"""
+struct CategoricalLikelihood end
+
+(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(softmax(append!(f, 0)))
```
devmotion, line 14: IMO we shouldn't silently mutate the input, that feels very unexpected to a user.

devmotion:
I'm not sure if @willtebbutt's intention was that https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/issues/18 should be addressed before merging any of the open likelihood PRs?

sharanry:
> I'm not sure if @willtebbutt's intention was that #18 should be addressed before merging any of the open likelihood PRs?

I thought he meant we address common tests before introducing anymore new likelihood PRs. I was waiting for these 2 PRs to get merged before starting on the TestUtils. Maybe I was mistaken.

devmotion:
> I thought he meant we address common tests before introducing anymore new likelihood PRs.

I assume that's what he had in mind, but since I'm not completely sure I think it's good if we wait and let him comment on it.

willtebbutt:
I'm not particularly fussed what order this happens in, provided that no new PRs are opened without having the test utils sorted :)

PR #17, opened 2020-09-17
Add Heteroscedastic Gaussian Likelihood

sharanry:


src/likelihoods/gaussian.jl
```diff
@@ -20,3 +20,23 @@ GaussianLikelihood() = GaussianLikelihood(1e-6)
...
+variance varies according to the input.
+
+```math
+    p(y|[f, σ]) = Normal(y | f, σ)
+```
+On calling, this would return a normal distribution with mean `f` and variance σ.
+"""
+struct HeteroscedasticGaussianLikelihood end
+
+@functor HeteroscedasticGaussianLikelihood
```
devmotion, line 38: No need for this since there are no parameters:
```suggestion

@functor HeteroscedasticGaussianLikelihood
```

src/likelihoods/gaussian.jl
```diff
@@ -20,3 +20,23 @@ GaussianLikelihood() = GaussianLikelihood(1e-6)
...
+
+"""
+    HeteroscedasticGaussianLikelihood(σ²)
+
+Heteroscedastic Gaussian likelihood. This is to be used if we assume that the 
+uncertainity associated with the data follows a Gaussian distribution where the 
+variance varies according to the input.
+
+```math
+    p(y|[f, σ]) = Normal(y | f, σ)
```
theogf, line 32: Could you rewrite this as `p(y|f,g) = Normal(y | f, exp(g))`
That leaves it clear that sigma is another gp

src/likelihoods/gaussian.jl
```diff
@@ -20,3 +20,23 @@ GaussianLikelihood() = GaussianLikelihood(1e-6)
...
+```
+On calling, this would return a normal distribution with mean `f` and variance σ.
+"""
+struct HeteroscedasticGaussianLikelihood end
+
+@functor HeteroscedasticGaussianLikelihood
+
+(::HeteroscedasticGaussianLikelihood)(f::AbstractVector{<:Real}) = Normal(f[1], exp(f[2]))
+
+(::HeteroscedasticGaussianLikelihood)(fs::AbstractVector) = MvNormal([f[1]  for f in fs], exp.([f[2]  for f in fs]))
```
theogf, line 42:
```suggestion
(::HeteroscedasticGaussianLikelihood)(fs::AbstractVector) = Product(Normal.(first.(fs), exp.(last.(fs))))
```

src/GPLikelihoods.jl
```diff
@@ -7,7 +7,7 @@ using Functors
 
 import Distributions
 
-export GaussianLikelihood, PoissonLikelihood
+export GaussianLikelihood, HeteroscedasticGaussianLikelihood, PoissonLikelihood
```
theogf, line 10:
```suggestion
export GaussianLikelihood, 
           HeteroscedasticGaussianLikelihood,
           PoissonLikelihood
```

src/likelihoods/gaussian.jl
devmotion: Actually, I think it is better to use `MvNormal` here. In general, one shouldn't use `Product` but `product_distribution` (similar to `truncated` vs `Truncated`) since the latter can be implemented in a more efficient way. Since it forwards to `MvNormal` anyway for a vector of normal distributions (see https://github.com/JuliaStats/Distributions.jl/blob/7eeee622565a8abb6540edae7ce5d42cd7586b5d/src/multivariate/product.jl#L62-L66) I'd suggest to use
```suggestion
(::HeteroscedasticGaussianLikelihood)(fs::AbstractVector) = MvNormal(first.(fs), exp.(last.(fs)))
```

sharanry:
@devmotion @willtebbutt @theogf Any objections to merge this?

src/likelihoods/gaussian.jl
```diff
@@ -20,3 +20,21 @@ GaussianLikelihood() = GaussianLikelihood(1e-6)
...
+    HeteroscedasticGaussianLikelihood(σ²)
+
+Heteroscedastic Gaussian likelihood. This is to be used if we assume that the 
+uncertainity associated with the data follows a Gaussian distribution where the 
+variance varies according to the input.
+
+```math
+    p(y|[f, g]) = Normal(y | f, exp(g))
+```
+On calling, this would return a normal distribution with mean `f` and variance σ.
```
devmotion, line 34:
```suggestion
On calling, this would return a normal distribution with mean `f` and variance `exp(g)`.
```

src/likelihoods/gaussian.jl
```diff
@@ -20,3 +20,21 @@ GaussianLikelihood() = GaussianLikelihood(1e-6)
 (l::GaussianLikelihood)(f::Real) = Normal(f, sqrt(l.σ²))
 
 (l::GaussianLikelihood)(fs::AbstractVector{<:Real}) = MvNormal(fs, sqrt(l.σ²))
+
+"""
+    HeteroscedasticGaussianLikelihood(σ²)
+
+Heteroscedastic Gaussian likelihood. This is to be used if we assume that the 
+uncertainity associated with the data follows a Gaussian distribution where the 
+variance varies according to the input.
```
willtebbutt, line 29:
```suggestion
Heteroscedastic Gaussian likelihood.
This is a Gaussian likelihood whose mean and the log of whose variance are functions of the
latent process.
```

test/likelihoods/gaussian.jl
```diff
@@ -11,3 +11,25 @@
...
     @test keys(Functors.functor(lik)[1]) == (:σ²,)
 end
+
+@testset "HeteroscedasticGaussianLikelihood" begin
+    rng = MersenneTwister(123)
+    gp = GP(IndependentMOKernel(SqExponentialKernel()))
+    IN_DIM = 3
+    OUT_DIM = 2 # one for the mean the other for the log-standard deviation
+    N = 10
+    x = [rand(rng, IN_DIM) for _=1:N]
```
willtebbutt, line 21:
```suggestion
    x = [rand(rng, IN_DIM) for _ in 1:N]
```

willtebbutt (approved):
This broadly looks good to me, and I'm happy to see it merged.

I think that it's important that we address the issue I've just opened (#18 ) before pushing forward with any more likelihoods though, or our tests are quickly going to become unmaintainable.

willtebbutt:
Also, patch bump please :)

PR #22, opened 2020-10-23
Add example with PalmerPenguins

sharanry:


examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,65 @@
+using PalmerPenguins, Plots
+const TABLE = PalmerPenguins.load()
+
+using DataFrames, Random
+df = dropmissing(DataFrame(TABLE))
+df[randperm(nrow(df)), :]
```
devmotion, line 6: This does not modify `df` although I guess you would like to permute the rows:
```suggestion
df = df[randperm(nrow(df)), :]
```

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,65 @@
+using PalmerPenguins, Plots
+const TABLE = PalmerPenguins.load()
+
+using DataFrames, Random
+df = dropmissing(DataFrame(TABLE))
```
devmotion, line 5: It doesn't seem very useful to create a constant that is used only once?
```suggestion
using DataFrames, Random
df = dropmissing(DataFrame(PalmerPenguins.load()))
```

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,65 @@
...
+f = LatentGP(GP(k), BernoulliLikelihood(), 0.001)
+fx = f(x_train)
+logpdf(fx, rand(fx))
+
+
+using EllipticalSliceSampling, Distributions
+function logp(params; x=x_train, y=y_train)
+    kernel = ScaledKernel(
+        KernelFunctions.transform(
+            Matern52Kernel(),
```
devmotion, line 40: I'm not sure if Matern52Kernel is a good choice here

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,65 @@
...
+            ScaleTransform(exp(params[1]))
+        ),
+        exp(params[2])
+    )
+    f = LatentGP(GP(k), BernoulliLikelihood(), 0.001)
```
devmotion, line 45:
```suggestion
    f = LatentGP(GP(kernel), BernoulliLikelihood(), 0.001)
```

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,65 @@
...
+first(df, 5)
+
+# Binary classification using Bernoulli likelihood - sex
+
+unique(df[:sex])
+
+
+using KernelFunctions, AbstractGPs, GPLikelihoods
+x = RowVecs(Array(df[:, [:bill_length_mm, :bill_depth_mm, :flipper_length_mm, :body_mass_g]]))
+y = [sex=="female" ? true : false for sex in df[:sex]]
```
devmotion, line 16:
```suggestion
y = df[:sex] .== "female"
```

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,65 @@
...
+)
+
+x_train, y_train = x[1:266], y[1:266]
+x_test, y_test = x[267:end], y[267:end]
+
+
+k = Matern52Kernel()
+f = LatentGP(GP(k), BernoulliLikelihood(), 0.001)
+fx = f(x_train)
+logpdf(fx, rand(fx))
```
devmotion, line 33: I'm not sure what you want to compute here. I assume you would rather want to base your evaluation on e.g. `x_test` and `y_test`? However, a general issue is that you can evaluate `logpdf` only for a named tuple of `y` and `f` (the outputs in the GP) but only `y` is available and `f` is a latent variable. So either you would have to sample one specific `f` from the GP or marginalize out `f` (which is usually not possible in closed form).

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,65 @@
...
+function logp(params; x=x_train, y=y_train)
+    kernel = ScaledKernel(
+        KernelFunctions.transform(
+            Matern52Kernel(),
+            ScaleTransform(exp(params[1]))
+        ),
+        exp(params[2])
+    )
+    f = LatentGP(GP(k), BernoulliLikelihood(), 0.001)
+    fx = f(x_train)
```
devmotion, line 46:
```suggestion
    fx = f(x)
```

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,65 @@
...
+    return logpdf(fx, rand(fx))
```
devmotion, line 47: Same as above, these computations should include `y` and not just evaluate a random sample. For the correct log probability here you have to marginalize out `f` (the outputs of the GP). However, for inference it should be sufficient to sample a latent vector (at least that was the case for LGCP that I worked with a while ago).

sharanry:
@devmotion Thanks for pointing out all those silly mistakes! I should taken closer look before submitting the PR.

examples/PalmerPenguins.jl
sharanry: > However, for inference it should be sufficient to sample a latent vector (at least that was the case for LGCP that I worked with a while ago).

Do you mean that I should just consider one sample latent vector like I have done now?

devmotion: I mean the setup here is a bit strange (and I'm not sure if it is correct at all). Usually you would perform Bayesian inference for both the latent vector `f(x_train)` and the parameters. If you are only interested in the posterior distribution of the parameters, preferably you marginalize out the latent vector if possible and only perform inference of the parameters (however, often this is not possible) - currently you use a MC approximation of it but I'm not sure if that's good enough. Additionally, to be able to use elliptical slice sampling, you have to ensure that the prior of the latent vector + the parameters is normally distributed.

In general it can be tricky to work with GPs with non-Gaussian likelihoods (see e.g. https://www.tandfonline.com/doi/pdf/10.1080/00949655.2017.1326117?needAccess=true or https://www.jmlr.org/papers/volume9/nickisch08a/nickisch08a.pdf). Often one also does not perform inference of both the parameters and the latent vector but optimizes the hyperparameters by maximizing the model evidence.

sharanry: Yeah I realize that. A naive way would probably be to fit the latent vectors (f) directly to the outputs (y). But this wouldn't be making use of the Bernoulli likelihood. I think section 3.4 from GPML presents another way to accomplish binary classification. I will look into all and get back to you.

sharanry: @devmotion I have tried to incorporate some of what you said. Could you take a second look and let me know if I am on the right track? Thanks!

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,125 @@
...
+        kernel = ScaledKernel(
+                KernelFunctions.transform(
+                    SqExponentialKernel(),
+                    ScaleTransform(exp(posterior_params[i,1]))
+                ),
+                exp(posterior_params[i,2])
+            )
+        p_fx = posterior(GP(kernel)(x_train), posterior_params[i,3:end])
+        l_p_fx = LatentGP(p_fx, BernoulliLikelihood(), 0.01)
+        (_, ys[i]) = rand(l_p_fx(x))
```
sharanry, line 95: @devmotion I am particularly doubtful if this part is right. Is this the right way to predict on test data?

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,125 @@
...
+        df[!, :bill_length_mm],
+        df[!, :bill_depth_mm],
+        df[!, :flipper_length_mm],
+        marker_z=mean_ys,
+        m = (0.5, :h, 5),
+        colorbar=:left
+    )
+
+# train data accuracy
+mean(((mean_ys[end-length(y_test)+1:end] .> 0.5) .== y_test))
```
sharanry, line 113: This given ~0.63 however test accuracy is ~0.37. We are currently overfitting.

examples/PalmerPenguins.jl
sharanry: I think this approximates conditioned posterior w.r.t to latent parameters.

devmotion: Basically, you have to compute (or approximate) eq. 3.9 and 3.10 in GPML where in 3.9 you additionally integrate over the kernel parameters. It seems you approximate 3.10 by Monte Carlo integration with respect to the kernel parameters and the latent vector. It is not clear how many samples are needed for a good approximation here but principally it is fine if you make sure that the samples from the MCMC method are reasonably good samples from the posterior. It would be good to inspect them and e.g. plot them and inspect statistics such as the effective sample size or rhat.

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,125 @@
...
+# train data accuracy w.r.t latent vector
+mean((mean_params[1,3:end] .> 0) .== y_train)
+
+ℓ(mean_params)
+
+plt = histogram(samples_mat[:, 1:2]; layout=2, labels= "Param", bins=10)
+vline!(plt, mean_params[:, 1:2]; layout=2, label="Mean")
+
+function posterior_mean(posterior_params)
+    ys = Vector(undef, size(posterior_params, 1))
```
devmotion, line 84: This leads to `Vector{Any}`. I would suggest to use `map` instead of a for loop here.

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,125 @@
...
+
+# +
+# (f_, y_) = rand(fx)
+# plt = plot()
+# plot!(plt, f_)
+# scatter!(plt, y_.-0.5)
+# -
+
+samples = sample(ESSModel(prior, ℓ), ESS(), 1_000; progress=true)
+samples_mat = reduce(hcat, samples)';
```
devmotion, line 71: Is it actually useful to transpose this matrix? In the posterior mean calculations it would be more efficient if each column would correspond to one sample.

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,125 @@
...
+)
+
+x_train, y_train = x[1:266], y[1:266]
+x_test, y_test = x[267:end], y[267:end];
+
+
+# logpdf without any parameters
+
+k = SqExponentialKernel()
+f = LatentGP(GP(k), BernoulliLikelihood(), 0.001)
```
devmotion, line 39: This is not consistent with what is used during inference:
```suggestion
f = LatentGP(GP(k), BernoulliLikelihood(), 0.01)
```

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,125 @@
...
+    for i in 1:size(posterior_params, 1)
+        kernel = ScaledKernel(
+                KernelFunctions.transform(
+                    SqExponentialKernel(),
+                    ScaleTransform(exp(posterior_params[i,1]))
+                ),
+                exp(posterior_params[i,2])
+            )
+        p_fx = posterior(GP(kernel)(x_train), posterior_params[i,3:end])
+        l_p_fx = LatentGP(p_fx, BernoulliLikelihood(), 0.01)
```
devmotion, line 94: This is also not consistent with inference:
```suggestion
        l_p_fx = LatentGP(p_fx, BernoulliLikelihood(), 0.1)
```

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,125 @@
...
+# scatter!(plt, y_.-0.5)
+# -
+
+samples = sample(ESSModel(prior, ℓ), ESS(), 1_000; progress=true)
+samples_mat = reduce(hcat, samples)';
+
+mean_params = mean(samples_mat; dims=1)
+
+# train data accuracy w.r.t latent vector
+mean((mean_params[1,3:end] .> 0) .== y_train)
```
devmotion, line 76: Shouldn't the train data accuracy be computed with the `posterior_mean` function below as well?

examples/PalmerPenguins.jl
devmotion: You have to evaluate the train accuracy in the same way for it to be comparable (see my comment above).

examples/PalmerPenguins.jl
sharanry: This was done mostly for the plotting function. I will try to find a work around.

examples/PalmerPenguins.jl
sharanry: The average ESS across parameters is pretty low -- ~7.5. Tuning the prior of ESS helped a little.

examples/PalmerPenguins.jl
sharanry: Did that. Thanks! There seems to be high variability in the accuracy. Probably due to terrible ESS.

examples/PalmerPenguins.jl
```diff
@@ -0,0 +1,135 @@
...
+        if Δ < 0
+            break
+        else
+            τ_inv += 2*Δ
+            K += 2
+        end
+    end
+    τ = min(1 / τ_inv, one(τ_inv))
+    τ * length(x)
+end
```
devmotion, line 87: IMO we should not reimplement existing functionality here but use MCMCDiagnostics or MCMCChains (implements different variants of ESS) directly.

sharanry: The idea was to avoid another dependency. But we might as well have this one as we are expected to use it frequently.

willtebbutt:
Closing because this is definitely stale, and hasn't seen any movement in a couple of years. Can re-open if someone wants to complete it

PR #23, opened 2020-10-24
Test reconstruction of parameters

devmotion:
BTW currently it is even more important to test the reconstructions in KernelFunctions for custom implementations of `Functors.functor` (this is needed if not all fields are parameters).


st-- (approved):
Looks good to me!

PR #38, opened 2021-08-25
Update construction MvNormal in Heteroscedastic

theogf:
Following a warning in the docs, the constructor of `MvNormal` has been changed
I also changed that the transform leads to the variance and not the std. dev. (which is more standard in the literature)
There was also a bug for the homoscedastic case (taking the std. dev. instead of the variance)

### DIFF
--- a/src/likelihoods/gaussian.jl
+++ b/src/likelihoods/gaussian.jl
@@ -21,19 +21,19 @@ GaussianLikelihood(σ²::Real) = GaussianLikelihood([σ²])
 
 (l::GaussianLikelihood)(f::Real) = Normal(f, sqrt(first(l.σ²)))
 
-(l::GaussianLikelihood)(fs::AbstractVector{<:Real}) = MvNormal(fs, sqrt(first(l.σ²)))
+(l::GaussianLikelihood)(fs::AbstractVector{<:Real}) = MvNormal(fs, first(l.σ²) * I)
 
 """
     HeteroscedasticGaussianLikelihood(l::AbstractLink=ExpLink())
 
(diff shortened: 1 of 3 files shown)


src/likelihoods/gaussian.jl
```diff
@@ -47,5 +47,5 @@ function (l::HeteroscedasticGaussianLikelihood)(f::AbstractVector{<:Real})
 end
 
 function (l::HeteroscedasticGaussianLikelihood)(fs::AbstractVector)
-    return MvNormal(first.(fs), l.invlink.(last.(fs)))
+    return MvNormal(first.(fs), l.invlink.(last.(fs)) * I)
```
devmotion, line 50: This should be:
```suggestion
    return MvNormal(first.(fs), Diagonal(l.invlink.(last.(fs)).^2))
```

devmotion (commented):
Can you update the homoscedastic case as well?

src/likelihoods/gaussian.jl
theogf: Actually it should be the opposite. We generally want to model the variance in heteroscedastic models, typically with `exp(f)`, `f^2`, or `var_max * logistic(f)`. One should probably change the one for the one-dimensional case

theogf: Oh but you're right about the Diagonal

devmotion (approved):
Looks good! Can you bump the version number to 0.2 before you merge it?

theogf:
Let's merge #40 first so we can get a first version of stable docs

PR #40, opened 2021-09-03
Add docs workflow

theogf:
Adds github action for generating docs


devmotion:
Maybe enable the preview in docs/make.jl and add a cleanup action as well?

PR #41, opened 2021-09-15
**BREAKING** swap order of parametrization of GammaLikelihood

st--:
To simplify dispatch on the type of invlink - which is what we'll normally want to dispatch on - whether \alpha is a Float32 or Float64 should generally be irrelevant.


theogf (approved):

willtebbutt:
@devmotion @st-- what's the status of this PR?

PR #44, opened 2021-09-28
Get rid of joinpath

theogf:
Similarly to KF.jl and GP.jl joinpath confuses the lintering and is not needed/recommended


devmotion (commented):
Can you also update https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/blob/master/test/runtests.jl and update the version number?

devmotion (approved):

Issue #45, opened 2021-09-28
AbstractLikelihood supertype


theogf:
Should we have a global abstract type for the `Likelihood` objects? That would be practical (at least for me)

PR #46, opened 2021-09-29
Add AbstractLikelihood type and use operatorname for docstrings

theogf:
Solves #45 by adding an `AbstractLikelihood` type and use `\mathrm` for the distribution names in math mode


devmotion (commented):
I prefer `\operatorname` (eg also used in Distributions IIRC) but I guess it is a matter of taste.

theogf:
No preference either so let's be consistent with Distributions.jl!

theogf:
@devmotion Can I merge this? I need it for some other package development :stuck_out_tongue:

devmotion (approved):

PR #48, opened 2021-10-08
Add `inverse` and simplify type structure

devmotion:
This PR implements the suggestion in https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/issues/43#issuecomment-920974828.

More concretely, if one defines `GPLikelihoods.inverse(::typeof(g)) = f` (probably one wants to define the reverse definition as well :wink:) then one gets `inv(::Link{typeof(g)}) isa Link{typeof(f)}` for free. This allows to define `LogLink`, `ExpLink` just as aliases of `Link`.

(As a hidden feature, in addition to existing links the PR also adds the suggestion in the linked comment for `log1pexp` and `logexpm1`, i.e., `Link(log1pexp)` and `Link(logexpm1)` are invertible now as well.)

In the future, maybe definitions for `inv(exp)` etc. might be available in base (https://github.com/JuliaLang/julia/issues/42421) or as type piracy in some package which would make this approach even more composable and put less maintenance burden on us.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -6,6 +6,12 @@ on:
       - master
   pull_request:
 
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
 jobs:
   test:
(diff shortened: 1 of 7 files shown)


theogf:
I have to say I find them quite need.
I'd rather see `Likelihood{<:LogisticLink}` then `Likelihood{<:Link{typeof(logistic)}}` (which also implies one would need to have `logistic` in the workspace.

oschulz:
ToDo: Switch to InverseFunctions. ;-)

devmotion:
There's already an issue 😉

We can't switch before LogExpFunctions and StatsFuns support it though (if we don't want to introduce type piracy).

Issue #43, opened 2021-09-16
Abandon `AbstractLink` type


st--:
All that I need to specify a likelihood is the mapping from f to a distribution, and in many cases the only degree of freedom is the inverse link function: it would be super convenient if I could just write `PoissonLikelihood(log1pexp)` to try out a softplus invlink, without having to code up a subtype of AbstractLink and having to remember what I have to do to implement it.

st--:
We could still dispatch analytic implementations of expected_loglik etc on `typeof(exp)` etc, right?

theogf:
How about creating a `GenericLink` and have
```
convert(::Type{AbstractLink}, x::Any) = GenericLink(x)
```
?

st--:
Just so I understand better, what do we gain from defining e.g. ExpLink explicitly (rather than "only" having GenericLink(exp))?

theogf:
We are able to tell what the corresponding inverse `Link` is `LogLink`

devmotion:
An alternative would be to add an API like
```julia
inverse_link(::typeof(exp)) = log
inverse_link(::typeof(log)) = exp

inverse_link(::typeof(log1pexp)) = logexpm1
inverse_link(::typeof(logexpm1)) = log1pexp
```
that would allow us to define
```julia
Base.inv(l::GenericLink) = GenericLink(inverse_link(l.f))
```

PR #51, opened 2021-10-08
Generalize likelihoods

devmotion:
This is a follow-up to #48.

Fixes the issue described in #43: now one can just use `PoissonLikelihood(log1pexp)` and everything will work as expected.


devmotion:
BTW sorry for the amount of commits, I noticed that it's quite challenging to make a PR with the webinterface on my phone without making stupid errors and separate commits for each file 😂

PR #53, opened 2021-10-14
Add basic docs

theogf:
Adds some very basic docs explaining what the package does


docs/src/index.md
```diff
@@ -1,8 +1,22 @@
...
+```@example
+    f = 2.0
+    GaussianLikelihood()(f) == Normal(2.0)
 ```
 
-```@autodocs
-Modules = [GPLikelihoods]
+Since the parameter domain of a lot of distributions is more restricted to the real
+domain, for example the parameter of a `Bernoulli` distribution, most `AbstractLikelihood`
+contain a `invlink` field which will map the latent variable to the right domain.
```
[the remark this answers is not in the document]
theogf: You mean the documentation of the concrete "likelihood" types right?

willtebbutt:
@theogf since we're getting PRs merged, what's the status of this PR?

theogf:
I think it needs another review pass, I made modifications following your comments and @st-- comments as well.

docs/src/api.md
```diff
@@ -0,0 +1,45 @@
...
+```@docs
+Link
+ChainLink
+```
+
+The rest of the links [`ExpLink`](@ref), [`LogisticLink`](@ref), etc.,
+are aliases for the corresponding wrapped functions in a `Link`.
+For example `ExpLink == Link{typeof(exp)}`.
+
+When passing a [`Link`](@ref) to an `AbstractLikelihood`, that this link
```
willtebbutt, line 29:
```suggestion
When passing a [`Link`](@ref) to an `AbstractLikelihood`, this link 
```

docs/src/index.md
```diff
@@ -1,8 +1,36 @@
-# GPLikelihoods.jl
+```@meta
+CurrentModule = GPLikelihoods
+```
+# GPLikelihoods
+
+[`GPLikelihoods.jl`](https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl) provides a practical interface to connect non-conjugate likelihoods
+with Gaussian Processes.
+The API is very basic: Every `AbstractLikelihood` object is a [functor](https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects-1)
+that takes the output of a Gaussian process as input and returns a
```
willtebbutt, line 9: Could we be more specific about what we mean by "output". Do we really just mean `Real`s by "output"?

willtebbutt (commented):
A couple of minor remarks.

docs/src/index.md
theogf: I see your point, I tried to be general since we accept both `Real` and `Vector` but this might be more confusing than anything. I will try to improve this part

PR #57, opened 2021-12-16
Move test interface in src

theogf:
I am trying to move the test interface into `src` and make it an internal module.
Obviously there is a dependency issue because of `KernelFunctions` and `AbstractGPs` but I thought that loading them externally would work anyway.

[EDIT:] So now the tests are simplified and do not depend on any GP constituents. Just what the input dimension should be.
I saved the old interface under `test/TestInferface.jl.save` if we want to move them into a different package (like `ApproximateGPs`) at some point.


devmotion:
> Obviously there is a dependency issue because of KernelFunctions and AbstractGPs but I thought that loading them externally would work anyway.

No, that won't work. But is it necessary to use KernelFunctions and/or AbstractGPs in the tests at all? The interface doesn't require them? To me it seems it would make more sense to test these integrations in the respective packages, i.e., KernelFunctions and AbstractGPs, or in separate tests in tests/...; but not as part of `test_interface`.

theogf:
> 

I agree. I could just remove these dependencies ( or maybe store them somewhere to be used in `ApproximateGPs` and such) and simplify the interface

Project.toml
```diff
@@ -10,10 +10,11 @@ InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 Distributions = "0.19, 0.20, 0.21, 0.22, 0.23, 0.24, 0.25"
 Functors = "0.1, 0.2"
 InverseFunctions = "0.1.2"
-StatsFuns = "0.9.13"
```
devmotion, line 18: Older versions of StatsFuns tend to cause problems with LogExpFunctions and IrrationalConstants - StatsFuns did not depend on LogExpFunctions and IrrationalConstants and therefore it is possible to load older versions of StatsFuns that already define the irrationals and additionally LogExpFunctions and IrrationalConstants which also define them. These leads to many redefinitions.

theogf: Alright I will put it back, I was trying this because of unrelated issues of incompatibilities because of `KernelFunctions` but that should not be a problem anymore

test/TestInterface.jl.save
```diff
@@ -0,0 +1,65 @@
+module TestInterface
```
willtebbutt, line 1: What's this file about?

theogf: To move these tests to ApproximateGPs or something like this. But I should probably just save it as a gist somewhere

willtebbutt: Ohhh okay. I can't say that I'm hugely fussed either way, provided it gets moved quickly.

theogf:
@willtebbutt @devmotion Could you approve the PR?

test/likelihoods/categorical.jl
```diff
@@ -4,9 +4,9 @@
     end
 
     lik = CategoricalLikelihood()
-    IN_DIM = 3
+    # IN_DIM = 3
```
willtebbutt, line 7:
```suggestion
```

test/likelihoods/categorical.jl
```diff
@@ -4,9 +4,9 @@
...
     OUT_DIM = 4
-    N = 10
-    X = MOInput([rand(IN_DIM) for _ in 1:N], OUT_DIM)
-    test_interface(lik, IndependentMOKernel(SqExponentialKernel()), X)
+    # N = 10
+    # X = MOInput([rand(IN_DIM) for _ in 1:N], OUT_DIM)
```
willtebbutt, line 10:
```suggestion
```

test/likelihoods/gaussian.jl
```diff
@@ -13,6 +13,6 @@ end
     IN_DIM = 3
     OUT_DIM = 2 # one for the mean the other for the log-standard deviation
     N = 10
-    X = MOInput([rand(IN_DIM) for _ in 1:N], OUT_DIM)
-    test_interface(lik, IndependentMOKernel(SqExponentialKernel()), X)
+    # X = MOInput([rand(IN_DIM) for _ in 1:N], OUT_DIM)
```
willtebbutt, line 16:
```suggestion
```

willtebbutt (approved):
I'm happy for this to be merged once the tidying-up related suggestions I've left have been dealt with. Sorry for the delay.

theogf:
No problem thanks for the quick reaction!

Issue #55, opened 2021-11-04
Categorical takes C-1 inputs for C classes


rossviljoen:
Currently, the `CategoricalLikelihood` is defined to take a vector of C-1 inputs to produce a distribution with C classes by appending a 0 to the input vector before going through the softmax.

This is fine for simple stuff, but I think there are some cases where you'd want to give all C inputs - e.g. if you're doing multi-class classification with a different kernel for each class, I don't think this would be possible with the current version?

Should I make a PR to change it or is there a reason to keep it as is? (would always still be possible to use the current version by appending a zero yourself)

@willtebbutt

devmotion:
The motivation for fixing one input was to ensure that the mapping is invertible: we map C-1 inputs to the C-1 dimensional simplex. It is the natural generalization of the logistic function, as used e.g. in [multinomial logistic regression](https://en.wikipedia.org/wiki/Multinomial_logistic_regression). I can see though that it can be a bit inconvenient sometimes.

rossviljoen:
What do you think then - change it or leave as is?

theogf:
I have been dealing with categorical likelihoods again recently and I think both are just as valid (interestingly having C inputs adds an unnecessary degree of freedom, and I am not sure what the effect on inference are).
I wanted to add that the choice of the added input (`0`) should be changeable by the user.

I will make a PR to allow for all these options, maybe I can find an elegant formulation.

theogf:
Solved by #61 I believe

theogf:
@devmotion Could you comment on the exchangability of the classes when using the C-1 inputs? Would it still be valid?

devmotion:
I'm not sure, what exactly do you mean?

theogf:
In the C inputs, C classes I can interchange any class by interchanging the input right.
But is it also true for C - 1 inputs.
In another words, is a simplex invariant under permutations?

devmotion:
If you interchange two of the C-1 inputs, then the probabilities of the corresponding two classes will be interchanged as well. And if you want to interchange some class with the reference class, you can either change the reference class or set its input to the additive inverse and subtract it from all other inputs. Is that what you're after?

[...]

PR #61, opened 2022-01-27
Rework categorical to allow multiple variants

theogf:
This solves #55 by having two variants (Simplex and Curved).
This add a new parametric type to `Categorical`. `SimplexVariant` (the original version) and `CurvedVariant` (the overparametrized version).
`SimplexVariant` is pretty straightforward to understand I believe (N-1 inputs for N categories)
`CurvedVariant` comes from the fact that such a parametrization leads to a "Curved Exponential Family" (see the small text after the table in https://en.wikipedia.org/wiki/Exponential_family#Table_of_distributions)

I am really open to other naming, especially the `CurvedVariant` one.
Maybe `OverparametrizedVariant` or something like this?


src/likelihoods/categorical.jl
```diff
@@ -1,22 +1,45 @@
...
+where it corresponds to Variant 1 and 2.
 """
-struct CategoricalLikelihood{Tl<:AbstractLink} <: AbstractLikelihood
+struct CategoricalLikelihood{Tb,Tl<:AbstractLink} <: AbstractLikelihood
     invlink::Tl
+    CategoricalLikelihood{Tb}(invlink::Tl) where {Tb,Tl} = new{Tb,Tl}(invlink)
 end
 
-CategoricalLikelihood(l=softmax) = CategoricalLikelihood(Link(l))
+function CategoricalLikelihood(l=softmax, bijective=true)
```
devmotion, line 29: Maybe make `bijective` a keyword argument? Type inference here relies on constant propagation, in general it won't be inferrable.

theogf: Wouldn't that be the advantage of having a singleton type `Bijective` ?

devmotion: You get type inference if you use e.g.
```suggestion
function CategoricalLikelihood(l=softmax; bijective=Val(true))
```
And in case the choice is fixed (which might often be the case), then constant propagation should work. But since it can fail sometimes, it would be good to support `Val` (possibly in addition to regular booleans).

An advantage of boolean values (types) would be that it clearly indicates there are only two choices.

devmotion: And, of course, you don't have to come up with names for singleton types and also don't have to export them.

theogf: Ok but forcing users to use `Val` is really weird IMO

devmotion: I'm used to it from SciML :smile: And you don't have to force it, you could support both.

theogf: And AFAIK we can't multiple dispatch on keywords...
But we could have
```
CategoricalLikelihood(l=softmax, bijective::Val)
CategoricalLikelihood(l=softmax; bijective=true)
```
?

devmotion: I thought you could use something like
```suggestion
function CategoricalLikelihood(l=softmax; bijective=Val(true))
    return CategoricalLikelihood{bijective_typeparameter(bijective)}(l)
end

bijective_typeparameter(bijective::Bool) = bijective
bijective_typeparameter(::Val{T}) = T
```

theogf:
That should be ready to merge, maybe we can merge #41 at the same time?

src/likelihoods/categorical.jl
```diff
@@ -1,22 +1,52 @@
...
+This variant is over-parametrized, as there are `n-1` independent parameters 
+embedded in a `n` dimensional parameter space.
+For more details, see the end of the section of this [Wikipedia link](https://en.wikipedia.org/wiki/Exponential_family#Table_of_distributions)
+where it corresponds to Variant 1 and 2.
 """
-struct CategoricalLikelihood{Tl<:AbstractLink} <: AbstractLikelihood
+struct CategoricalLikelihood{Tb,Tl<:AbstractLink} <: AbstractLikelihood
     invlink::Tl
+    CategoricalLikelihood{Tb}(invlink) where {Tb} = CategoricalLikelihood{Tb}(Link(invlink))
+    CategoricalLikelihood{Tb}(invlink::Tl) where {Tb,Tl<:AbstractLink} = new{Tb,Tl}(invlink)
```
devmotion, line 28: Alternatively, one could use
```suggestion
    function CategoricalLikelihood{Tb}(invlink) where {Tb}
        l =  Link(invlink)
        new{Tb,typeof(l)}(l)
    end
```

devmotion (commented):
Maybe you could add some `@inferred` tests?

src/likelihoods/categorical.jl
theogf: But with your proposition if one passes a `Link` like `SoftMaxLink` it's going to be wrapped twice.

theogf:
> Maybe you could add some `@inferred` tests?

I am sorry but I am not sure how it would work. I should just have lines like
```
@inferred CategoricalLikelihood(args...; kwargs...)
```
?

theogf:
Ah I see now, @inferred will fail with `Bool` arguments but pass with `Val{true/false}`

src/likelihoods/categorical.jl
theogf: This issue is dealt with in #65

src/likelihoods/categorical.jl
```diff
@@ -1,22 +1,54 @@
...
-    CategoricalLikelihood(l=softmax)
+    CategoricalLikelihood(l=softmax; bijective::Union{Bool,Val{false},Val{true}}=Val(true))
 
 Categorical likelihood is to be used if we assume that the 
-uncertainity associated with the data follows a Categorical distribution.
+uncertainty associated with the data follows a [Categorical distribution](https://en.wikipedia.org/wiki/Categorical_distribution) with `n` categories.
 ```math
     p(y|f_1, f_2, \\dots, f_{n-1}) = \\operatorname{Categorical}(y | l(f_1, f_2, \\dots, f_{n-1}, 0))
+    p(y|f_1, f_2, \\dots, f_n) = \\operatorname{Categorical}(y | l(f_1, f_2, \\dots, f_n))
 ```
```
devmotion, line 9: Maybe move the formulas to the corresponding sections below?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "GPLikelihoods"
 uuid = "6031954c-0455-49d7-b3b9-3e1c99afaf40"
 authors = ["JuliaGaussianProcesses Team"]
-version = "0.2.7"
+version = "0.3.0"
```
devmotion, line 4: Maybe we could make the breaking change of the `GammaLikelihood` in the same release (not sure if the PR is ready)?

devmotion (approved):
LGTM, only minor questions/suggestions left.

theogf:
@devmotion 
Somehow I am not happy with the current setup.
- First the first parameter of `CategoricalLikelihood` is not an `AbstractLink` (as we try to enforce for most likelihoods right now)
- Second, I feel that it should be the `Link` responsible for deciding the variant (the likelihood itself is not really bijective)

How about instead having the variants included as a wrapper for the `Link`s, like
```julia
struct BijectiveSimplexLink{L} <: AbstractLink
    l::L
end

(l::BijectiveSimplexLink)(f::AbstractVector{<:Real}) = l(vcat(f, 0))
# etc...
```

devmotion:
I'm not so much worried about type parameters (one could also rearrange them) but I think the second point is valid. However, I wonder if all these considerations are valid/relevant for anything else than the `softmax` function. Maybe it would be even more "correct" to just define two softmax links, one that is bijective and one that isn't. And make the bijective one the default but mention the other one in the docstring. And don't change the `Categorical...` type at all?

theogf:
> all these considerations are valid/relevant for anything else than the softmax function

There are already two other links in this situation the "Heaviside" one (typically `i == findmax(f)`) or from my own work the `logisticsoftmax` where instead of `exp` you use `logistic`. This would nicely wrap around those.

devmotion:
I see, in this case it seems reasonable to define the link types in the general fashion you outlined and allow other wrapped links besides `softmax`. But even in the general case we should only change the default link but not the `Categorical...` type with this design, I think.

src/likelihoods/categorical.jl
```diff
@@ -1,22 +1,46 @@
...
 struct CategoricalLikelihood{Tl<:AbstractLink} <: AbstractLikelihood
     invlink::Tl
 end
 
-CategoricalLikelihood(l=softmax) = CategoricalLikelihood(link(l))
+function CategoricalLikelihood(
+    l=softmax; bijective::Union{Bool,Val{true},Val{false}}=Val(true)
+)
+    return CategoricalLikelihood(make_bijective(link(l), bijective))
+end
```
devmotion, line 38: What I meant: I don't think we should not introduce `bijective` at all - since it makes it more difficult to use a custom user-defined link function. Even though you code everything yourself you would have to deal with `bijective` and end up with a wrapper that's not needed.

More concretely, I thought something like this should be sufficient:
```julia
CategoricalLikelihood(l=BijectiveSimplexLink(softmax)) = CategoricalLikelihood(link(l))

(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(l.invlink(f))

function (l::CategoricalLikelihood)(fs::AbstractVector)
    return Product(Categorical.(l.invlink.(fs)))
end

struct BijectiveSimplexLink{T}
    link::T
end

(l::BijectiveSimplexLink)(f::AbstractVector{<:Real}) = l.link(vcat(f, 0))
```
```

PR #62, opened 2022-01-28
Correct typo in `LogisticLink` docstring

simsurace:
The formula should read exp(f)/(1+exp(f)) or 1/(1+exp(-f)). I chose the latter for simplicity.


theogf:
You're right thanks for the correction!

theogf (approved):

PR #63, opened 2022-01-28
Add negative binomial likelihood

simsurace:
I opted for the p = logistic(f) parameterization. I'm not quite sure whether there are any tradeoffs in choosing this vs. e.g. \mu = exp(f), where \mu is the conditional mean of the negative binomial distribution. The latter parametrization leads to the mean to be only affected by the variance and mean of the latent GP, and not by the likelihood parameter. I haven't thought about whether this has any benefits for parameter learnability that might make it preferable to the p=logistic(f) parameterization.


src/likelihoods/negativebinomial.jl
```diff
@@ -0,0 +1,26 @@
...
+
+Negative binomial likelihood with number of successes `r`.
+
+```math
+    p(k|f) = \\operatorname{NB}(k | r, l(f))
+```
+On calling, this would return a negative binomial distribution with `r` successes and probability of success equal to `l(f)`.
+Note the distinction of the role of parameter `r` to the Wikipedia definition.
+"""
+struct NegBinomialLikelihood{T<:Real,Tl<:AbstractLink} <: AbstractLikelihood
```
devmotion, line 12: Maybe we could use the name `NegativeBinomialLikelihood`, to be consistent with the name of the distribution? I guess also it's more common to dispatch on the link, so
```suggestion
struct NegativeBinomialLikelihood{Tl<:AbstractLink,T<:Real} <: AbstractLikelihood
```
might be more useful.

src/likelihoods/negativebinomial.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    NegBinomialLikelihood(r::Real=1.0, l=logistic)
+
+Negative binomial likelihood with number of successes `r`.
+
+```math
+    p(k|f) = \\operatorname{NB}(k | r, l(f))
```
theogf, line 7: Could you write the explicit formulation of the pdf. The `NegativeBinomial` has multiple formulations and it would good to clarify it.

simsurace: Sure!

src/likelihoods/negativebinomial.jl
simsurace: I think this makes sense.

src/likelihoods/negativebinomial.jl
```diff
@@ -0,0 +1,26 @@
...
+On calling, this would return a negative binomial distribution with `r` successes and probability of success equal to `l(f)`.
+Note the distinction of the role of parameter `r` to the Wikipedia definition.
+"""
+struct NegBinomialLikelihood{T<:Real,Tl<:AbstractLink} <: AbstractLikelihood
+    r::T    # number of failures parameter
+    invlink::Tl
+end
+
+NegBinomialLikelihood(l) = NegBinomialLikelihood(1.0, l)
+NegBinomialLikelihood(r::Real=1.0, l=logistic) = NegBinomialLikelihood(r, Link(l))
```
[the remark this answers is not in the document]
theogf: Also now that #65 has been merged this should be written as 
```suggestion
NegBinomialLikelihood(l=logistic; r::Real=1) = NegBinomialLikelihood(r, link(l))
```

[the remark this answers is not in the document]
simsurace: Why not? But would this be pursposefully different from `GammaLikelihood`?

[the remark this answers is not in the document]
theogf: Actually we are trying to solve this for the `GammaLikelihood` already in #41

src/likelihoods/negativebinomial.jl
simsurace: Do you suggest a type parameter order different from e.g. `GammaLikelihood` on purpose?

devmotion: Yes, since we are not happy with the order in `GammaLikelihood`: https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/pull/41

src/likelihoods/negativebinomial.jl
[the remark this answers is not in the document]
simsurace: Oh I see.

simsurace:
Apparently the keyword argument constructor breaks the test for when `logistic` is not being wrapped in the `link`. 
How shall we resolve this? Also introduce the keyword constructor for `GammaLikelihood` and change the tests?

devmotion:
Change the tests similar to https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/pull/61/files#diff-d19095b0c530c06b29a86657ed746b5d40a0ebf453a75de8b02aa31b6acb119f, I'd suggest. In particular, it would be good to add tests with the keyword argument and to use also non-default values (e.g., different values for `r`).

src/likelihoods/negativebinomial.jl
```diff
@@ -0,0 +1,25 @@
+"""
+    NegativeBinomialLikelihood(r::Real=1.0, l=logistic)
```
devmotion, line 2:
```suggestion
    NegativeBinomialLikelihood(l=logistic; r::Real=1)
```

devmotion: Maybe one could also use a more descriptive name for `r`? `successes`?

src/likelihoods/negativebinomial.jl
```diff
@@ -0,0 +1,25 @@
...
+
+Negative binomial likelihood with number of successes `r`.
+
+```math
+    p(k|f) = \\frac{\\Gamma(k+r)}{k! \\Gamma(r)} p^r (1 - p)^k
+```
+On calling, this would return a negative binomial distribution with `r` successes and probability of success equal to `l(f)`.
```
devmotion, line 9:
```suggestion
On calling, this returns a negative binomial distribution with `r` successes and probability of success equal to `l(f)`.
```

simsurace: Agree, I'll also change the analogous word in gamma.jl.

src/likelihoods/negativebinomial.jl
```diff
@@ -0,0 +1,25 @@
+"""
+    NegativeBinomialLikelihood(l=logistic; successes::Real=1)
+
+Negative binomial likelihood with number of successes `successes`.
+
+```math
+    p(k|f) = \\frac{\\Gamma(k+successes)}{k! \\Gamma(successes)} l(f)^successes (1 - l(f))^k
+```
+On calling, this returns a negative binomial distribution with `successes` successes and probability of success equal to `l(f)`.
+Note the distinction of the role of parameter `successes` to the parameter `r` in the Wikipedia definition.
```
theogf, line 10: Maybe you can write that to get the equivalent of the Wikipedia definition, one just needs to define p = 1 - p

src/likelihoods/negativebinomial.jl
```diff
@@ -0,0 +1,25 @@
+"""
+    NegativeBinomialLikelihood(l=logistic; successes::Real=1)
+
+Negative binomial likelihood with number of successes `successes`.
```
theogf, line 4: I come a bit late, but I disagree with the use of `successes` literally no other distributions has such explicit naming.

src/likelihoods/negativebinomial.jl
simsurace: Maybe this is slightly awkward, but the reason this is different from the Wikipedia definition is because Distributions.jl chose to use a different convention. I wonder whether it is better to be consistent with Wikipedia or within the Julia ecosystem.

src/likelihoods/negativebinomial.jl
simsurace: Happy to revert, let me know when you reach a consensus.

src/likelihoods/negativebinomial.jl
theogf: It's a very annoying one. I don't know why Distributions.jl went with this parameterization, maybe it's inherited from R...
Anyway I think we should be consistent with Distributions.jl but we just need to have a warning in big red letters in the docs and how to convert to the Wikipedia parameterization.

src/likelihoods/negativebinomial.jl
theogf: @devmotion what is the motivation behind the verbose argument? To make it clear it's different from Wikipedia?

theogf: @devmotion what is the motivation behind the verbose argument? To make it clear it's different from Wikipedia?

devmotion: Mainly to make it clear what the parameter means, similar to how we renamed some keyword arguments in KernelFunctions - `r` is not very descriptive I'd say 😅 And, of course, there's an additional source of confusion here due to the discrepancy with Wikipedia that could maybe be resolved by a more descriptive name.

theogf: Should we just do what we did for KF.jl, i.e. having 2 keyword arguments for the same thing as `... ; r, successes=r)` ?

theogf: > `r` is not very descriptive,

Just as mu and sigma yet everyone uses that. I think it's a matter of being familiar with the negative binomial distribution.

src/likelihoods/negativebinomial.jl
simsurace: Shall I add a boldface **CAUTION** to the docstring? On a related note, any thoughts about the initial comment in this PR?

src/likelihoods/negativebinomial.jl
devmotion: > > `r` is not very descriptive,
> 
> Just as mu and sigma yet everyone uses that. I think it's a matter of being familiar with the negative binomial distribution. 

I don't think anyone uses `mu` and `sigma` as parameters of a normal distribution for anything else than mean and standard deviation. But apparently there are different interpretations of `r` in a negative binomial distribution, as Wikipedia and Distributions show.

theogf: How about going crazy and starting here :)

simsurace: While I don't particularly care either way, IMHO users will usually consult the documentation when playing around with different distributions (even just to check which distributions are actually supported), so they will likely find out what `r` means.

simsurace: Plus, for end users who are fitting a `LatentGP` with negative binomial likelihood, actually it does not matter that much what the parameters mean until they are plotting the results. In that case, one could define some pretty printing facilities that include a parameter table with some descriptors.

src/likelihoods/negativebinomial.jl
simsurace: > Maybe you can write that to get the equivalent of the Wikipedia definition, one just needs to define p = 1 - p

Yes, I could add that, but note that at least for the logistic link this corresponds to an unidentifiable sign change of the latent GP.

devmotion: > Shall I add a boldface CAUTION to the docstring?

You can add a `!!! warning` admonition, this will show up prominently in the docs and the REPL: https://juliadocs.github.io/Documenter.jl/stable/showcase/#Admonitions

src/likelihoods/negativebinomial.jl
devmotion: > While I don't particularly care either way, IMHO users will usually consult the documentation when playing around with different likelihoods (even just to check which likelihoods are actually supported), so they will likely find out what r means.

IMO the main motivation is not to simplify coding since, as you say, usually one has to consult the documentation anyway. I think the main advantage of more descriptive arguments is the same as for e.g. descriptive function and variable names: it makes it easier to (re-)read the code and to reason about it.

theogf: Alright, alright, let's stick with `successes`, at least it will force people using `r=` to check the docs and see the warning

PR #64, opened 2022-01-28
Add a few words about the role of this package

simsurace:
I hope I got your intentions right.


theogf (approved):
Sounds good to me but I would like to also have the opinion of others

PR #65, opened 2022-01-28
Use smart `link` constructor

theogf:
Right now all likelihood constructor are wrapping passed `AbstractLink` in another `Link`. Of course it still works but it adds an unnecessary layer of complexity.
I added the `link` smart constructor which wraps functions and functors in a `Link` and is the identity when passed an `AbstractLink`.

### DIFF
--- a/src/GPLikelihoods.jl
+++ b/src/GPLikelihoods.jl
@@ -14,7 +14,6 @@ export BernoulliLikelihood,
     PoissonLikelihood,
     ExponentialLikelihood,
     GammaLikelihood
-export link
 export Link,
     ChainLink,
     ExpLink,
--- a/src/GPLikelihoods.jl
+++ b/src/GPLikelihoods.jl
@@ -14,6 +14,7 @@ export BernoulliLikelihood,
     PoissonLikelihood,
     ExponentialLikelihood,
     GammaLikelihood
+export link
 export Link,
     ChainLink,
     ExpLink,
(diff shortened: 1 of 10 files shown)


devmotion:
I was convinced that is how `Link` works 😮😮

src/GPLikelihoods.jl
```diff
@@ -14,6 +14,7 @@ export BernoulliLikelihood,
     PoissonLikelihood,
     ExponentialLikelihood,
     GammaLikelihood
+export link
```
devmotion, line 17: Maybe not necessary to export it? I guess it's mainly useful internally.

```suggestion
```

theogf: Yup I realized this later

theogf:
> I was convinced that is how `Link` works open_mouthopen_mouth

This would have been type unstable though no?

devmotion:
Just a bit misleading, no? I don't think there would have been a type stability issue. But it's cleaner this way :slightly_smiling_face:

devmotion (approved):
LGTM, can you bump the version number?

theogf:
> Just a bit misleading, no? I don't think there would have been a type stability issue. But it's cleaner this way slightly_smiling_face

I meant that if `Link(l::ChainLink)` would return a `ChainLink` this would be a problem no? Maybe it's not type stability but there is another name for this?

devmotion:
Yes, that's what I meant with misleading. It's violating a common convention but otherwise it's fine. I don't know a specific name for it though.

PR #66, opened 2022-01-31
Add BijectiveSimplexLink in the docs

theogf:


src/links.jl
```diff
@@ -36,10 +36,10 @@ Base.inv(l::Link) = Link(InverseFunctions.inverse(l.f))
...
 the link `link`.
 This is a necessary step to work with simplices.
-For example with the [`SoftMaxLink`](@ref), to obtain a `n`-simplex leading to
-`n+1` categories for the [`CategoricalLikelihood`](@ref),
-one needs to pass `n+1` latent GP.
-However, by wrapping the link into a `BijectiveSimplexLink`, only `n` latent are needed. 
+For example with the [`SoftMaxLink`](@ref), to obtain a `n-1`-simplex leading to
+`n` categories for the [`CategoricalLikelihood`](@ref),
+one needs to pass `n` latent GP.
+However, by wrapping the link into a `BijectiveSimplexLink`, only `n-1` latent GP are needed.
```
st--, line 42:
```suggestion
However, by wrapping the link into a `BijectiveSimplexLink`, only `n-1` latent GPs are needed. 
```

st--, line 42: What effect does this reparametrisation have on the model? Does it break the symmetry amongst classes (i.e., would changing the order of classes change the resulting fit)? Would it be useful to mention the 0 added at the end in the docstring, or is that irrelevant?

Also, a harder question that you may not have the answer to but I would be curious if this makes it easier to fit the model (because there's no more redundancy through the overall level, and hence it becomes identifiable)...

theogf: So in some quick 1-D experiments with a few classes I did in my thesis. I generated data via the [fs, 0] process and trying to recover the latent GPs with the bijective and non-bijective link.
I consistently observed that the bijective link produced more correct probability distributions (compared to the true generating probabilities) but that the non-bijective likelihood had a better log-likelihood on the training data

theogf: This was with the logistic-softmax link. Also the augmentation non-bijective creates improper priors whereas in the bijective case all is nice and beautiful.
In terms of speed there does not seem to be a difference, but I did not check thoroughly

theogf:
@st-- Did I address your comments?

theogf:
Since this is just a minor doc change, I think it can safely be merged?

Issue #58, opened 2021-12-16
`CategoricalLikelihood` should contain the number of outputs


theogf:
I am planning to write an `nlatent` function for the package, which is very practical to initialize things when given a likelihood.

However, the current version of `CategoricalLikelihood` has no way of knowing the number of latent needed. what do you think about enforcing the number of outputs in the likelihood?

willtebbutt:
I think I would like to see an example of how using this would look with / without this requirement. i.e. what would we currently do, vs what would we do if this data were stored in the likelihood.

theogf:
I guess something like:
```
struct CategoricalLikelihood{Tl<:AbstractLink} <: AbstractLikelihood
    nclass::Int
    invlink::Tl
end

CategoricalLikelihood(nclass, l=BijectiveSimplexLink(softmax)) = CategoricalLikelihood(nclass, link(l))

nlatent(l::CategoricalLikelihood) = l.nclass
nlatent(l::CategoricalLikelihood{<:BijectiveSimplexLink}) = l.nclass - 1
```

Eventually this could be used to check the input dimensionality. But I am more imagining a front end model where you would pass, kernel, likelihood and inference and that it could be automatically inferred how many latent GPs you need

theogf:
I will just do a PR proposal and we can see if it makes sense

willtebbutt:
Sorry, yeah, that seems like a good idea.

PR #68, opened 2022-03-10
Adds nlatent and nclass to Categorical

theogf:
As discussed in #58 I am adding a `nlatent` function that indicates how many latent GPs are neeeded by the likelihood.
It also means that `CategoricalLikelihood` needs to know how many classes are represented.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "GPLikelihoods"
 uuid = "6031954c-0455-49d7-b3b9-3e1c99afaf40"
 authors = ["JuliaGaussianProcesses Team"]
-version = "0.3.1"
+version = "0.3.2"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
--- a/src/GPLikelihoods.jl
+++ b/src/GPLikelihoods.jl
@@ -28,12 +28,24 @@ export Link,
     ProbitLink,
     NormalCDFLink,
     SoftMaxLink
+export nlatent
 
 # Links
 include("links.jl")
 
 # Likelihoods
 abstract type AbstractLikelihood end
+
+"""
+    nlatent(::AbstractLikelihood)::Int
+
+Returns the number of latent Gaussian processes needed to build the likelihood.
+In other terms the input dimensionality passed to the likelihood from the GP perspective.
+It is typically 1, but for some likelihoods like [`CategoricalLikelihood`](@ref) or 
+[`HeteroscedasticGaussianLikelihood`](@ref) multiple latent GPs are necessary.
+"""
+nlatent(::AbstractLikelihood) = 1 # Default number of latent GPs required is 1
+
 include("likelihoods/bernoulli.jl")
 include("likelihoods/categorical.jl")
 include("likelihoods/gaussian.jl")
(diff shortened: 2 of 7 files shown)


devmotion:
Isn't this information known anyway when you model something? Ie, if you choose to work with a heteroscedastic Gaussian likelihood you should know that you need two latent GPs?

theogf:
> Isn't this information known anyway when you model something? Ie, if you choose to work with a heteroscedastic Gaussian likelihood you should know that you need two latent GPs?

Right sure you do. But the motivation is to be able to know this information when building the model itself without having to specify it explicitly.

If we build later an API where we get something like `build_model(X, y, kernel, likelihood)` one can infer what kind of underlying GP structure we need just from the likelihood.

The reason I need this at the moment is for AugmentedGPLikelihoods.jl where I need to specify the initial size of the augmented variables based on the likelihood : see https://github.com/JuliaGaussianProcesses/AugmentedGPLikelihoods.jl/blob/d976cc0a37639872a520d3818f44f566c58d39f6/src/likelihoods/categorical.jl#L20

theogf:
I also want to mention that in AugmentedGPs.jl I initially had the categorical likelihood implemented such that it would directly infer the number of classes and others and this proved to be a bad decision. It's incompatible with the online setting, it fails when not all classes are present in a treated minibatch and others.
I ended up forcing to give information to the likelihood, the minimum being how many classes are present.

devmotion:
Hmm it's still not really clear to me why this information should be encoded in the likelihood, in particular for the categorical likelihood. Currently you can evaluate it with a latent vector of arbitrary size and it will return a categorical distribution with the corresponding number of classes automatically. I guess I'm just not familiar enough with your use case and my confusion arises from the fact that I don't see how one would build a model without knowing the number of GPs, classes, etc. already.

theogf:
Maybe let me give a very precise example where this would matter.
Assume we want to do multi-class classification in an online setting.
You **need** to know how many classes are there in advance.
Because even if you planned to infer that from a first batch, there is no guarantee that all classes are present in this batch and you might end up with the wrong number of latent GPs.
So, as you said also, we need to know the number of classes in advance, and it seems a natural thing to me to incorporate this information directly in the likelihood.

devmotion:
Thanks for the clear example! Now the point I don't understand is: why would want to encode this information in the likelihood if it works fine without specifying the number of classes? If you know the number of classes and you know you want to use a categorical link, with a bijective or a non-bijective link, then you know the number of latent GPs already? And only for this point, but not the likelihood, the number of classes is needed.

Or, put differently: You have to know the number of classes, and hence latent GPs, when coding your model. So why not just use this information where it matters, ie when coding the GP?

theogf:
Ok I think our discussion here resumes to, when creating a model, 
- The number of latent be inferrable from the likelihood(s) itself (my point of view).
- The number of latent should be provided by the user and the number of categories should be inferred from it (depending on the link as well) (your point of view if I get it correctly)

My argument for my POV is that it puts less burden on the user in the case of a high-end interface.
Give me a likelihood, give me some data and I will automatically build the right latent model. This allows to have a very general interface no matter what the likelihood is (and if need multiple latent GPs or not)

theogf:
I am just going to close this PR as it went stale.

PR #72, opened 2022-03-25
less code duplication

st--:
Less chance for mistakes / less work when changing anything later on.:)

### DIFF
--- a/src/likelihoods/bernoulli.jl
+++ b/src/likelihoods/bernoulli.jl
@@ -18,4 +18,4 @@ BernoulliLikelihood(l=logistic) = BernoulliLikelihood(link(l))
 
 (l::BernoulliLikelihood)(f::Real) = Bernoulli(l.invlink(f))
 
-(l::BernoulliLikelihood)(fs::AbstractVector{<:Real}) = Product(l.(fs))
+(l::BernoulliLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
--- a/src/likelihoods/bernoulli.jl
+++ b/src/likelihoods/bernoulli.jl
@@ -18,4 +18,4 @@ BernoulliLikelihood(l=logistic) = BernoulliLikelihood(link(l))
 
 (l::BernoulliLikelihood)(f::Real) = Bernoulli(l.invlink(f))
 
-(l::BernoulliLikelihood)(fs::AbstractVector{<:Real}) = Product(Bernoulli.(l.invlink.(fs)))
+(l::BernoulliLikelihood)(fs::AbstractVector{<:Real}) = Product(l.(fs))
--- a/src/likelihoods/categorical.jl
+++ b/src/likelihoods/categorical.jl
@@ -36,6 +36,4 @@ function (l::CategoricalLikelihood)(f::AbstractVector{<:Real})
     return Categorical(l.invlink(f))
 end
 
-function (l::CategoricalLikelihood)(fs::AbstractVector)
-    return Product(Categorical.(l.invlink.(fs)))
-end
+(l::CategoricalLikelihood)(fs::AbstractVector) = Product(l.(fs))
--- a/src/likelihoods/categorical.jl
+++ b/src/likelihoods/categorical.jl
@@ -36,4 +36,4 @@ function (l::CategoricalLikelihood)(f::AbstractVector{<:Real})
     return Categorical(l.invlink(f))
 end
 
-(l::CategoricalLikelihood)(fs::AbstractVector) = Product(l.(fs))
+(l::CategoricalLikelihood)(fs::AbstractVector) = Product(map(l, fs))
--- a/src/likelihoods/exponential.jl
+++ b/src/likelihoods/exponential.jl
@@ -15,4 +15,4 @@ ExponentialLikelihood(l=exp) = ExponentialLikelihood(link(l))
 
 (l::ExponentialLikelihood)(f::Real) = Exponential(l.invlink(f))
 
-(l::ExponentialLikelihood)(fs::AbstractVector{<:Real}) = Product(l.(fs))
+(l::ExponentialLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
--- a/src/likelihoods/exponential.jl
+++ b/src/likelihoods/exponential.jl
@@ -15,6 +15,4 @@ ExponentialLikelihood(l=exp) = ExponentialLikelihood(link(l))
 
 (l::ExponentialLikelihood)(f::Real) = Exponential(l.invlink(f))
 
-function (l::ExponentialLikelihood)(fs::AbstractVector{<:Real})
-    return Product(Exponential.(l.invlink.(fs)))
-end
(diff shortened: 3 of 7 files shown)


src/likelihoods/bernoulli.jl
```diff
@@ -18,4 +18,4 @@ BernoulliLikelihood(l=logistic) = BernoulliLikelihood(link(l))
 
 (l::BernoulliLikelihood)(f::Real) = Bernoulli(l.invlink(f))
 
-(l::BernoulliLikelihood)(fs::AbstractVector{<:Real}) = Product(Bernoulli.(l.invlink.(fs)))
+(l::BernoulliLikelihood)(fs::AbstractVector{<:Real}) = Product(l.(fs))
```
devmotion, line 21: Could just use `map` now ans avoid the broadcasting machinery since broadcast fusion is not used anymore:

```suggestion
(l::BernoulliLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
```

src/likelihoods/categorical.jl
```diff
@@ -36,6 +36,4 @@ function (l::CategoricalLikelihood)(f::AbstractVector{<:Real})
     return Categorical(l.invlink(f))
 end
 
-function (l::CategoricalLikelihood)(fs::AbstractVector)
-    return Product(Categorical.(l.invlink.(fs)))
-end
+(l::CategoricalLikelihood)(fs::AbstractVector) = Product(l.(fs))
```
devmotion, line 39:
```suggestion
(l::CategoricalLikelihood)(fs::AbstractVector) = Product(map(l, fs))
```

src/likelihoods/exponential.jl
```diff
@@ -15,6 +15,4 @@ ExponentialLikelihood(l=exp) = ExponentialLikelihood(link(l))
 
 (l::ExponentialLikelihood)(f::Real) = Exponential(l.invlink(f))
 
-function (l::ExponentialLikelihood)(fs::AbstractVector{<:Real})
-    return Product(Exponential.(l.invlink.(fs)))
-end
+(l::ExponentialLikelihood)(fs::AbstractVector{<:Real}) = Product(l.(fs))
```
devmotion, line 18:
```suggestion
(l::ExponentialLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
```

src/likelihoods/gamma.jl
```diff
@@ -20,4 +20,4 @@ GammaLikelihood(α::Real=1.0, l=exp) = GammaLikelihood(α, link(l))
 
 (l::GammaLikelihood)(f::Real) = Gamma(l.α, l.invlink(f))
 
-(l::GammaLikelihood)(fs::AbstractVector{<:Real}) = Product(Gamma.(l.α, l.invlink.(fs)))
+(l::GammaLikelihood)(fs::AbstractVector{<:Real}) = Product(l.(fs))
```
devmotion, line 23:
```suggestion
(l::GammaLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
```

src/likelihoods/negativebinomial.jl
```diff
@@ -33,6 +33,4 @@ end
 
 (l::NegativeBinomialLikelihood)(f::Real) = NegativeBinomial(l.successes, l.invlink(f))
 
-function (l::NegativeBinomialLikelihood)(fs::AbstractVector{<:Real})
-    return Product(NegativeBinomial.(l.successes, l.invlink.(fs)))
-end
+(l::NegativeBinomialLikelihood)(fs::AbstractVector{<:Real}) = Product(l.(fs))
```
devmotion, line 36:
```suggestion
(l::NegativeBinomialLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
```

src/likelihoods/poisson.jl
```diff
@@ -18,4 +18,4 @@ PoissonLikelihood(l=exp) = PoissonLikelihood(link(l))
 
 (l::PoissonLikelihood)(f::Real) = Poisson(l.invlink(f))
 
-(l::PoissonLikelihood)(fs::AbstractVector{<:Real}) = Product(Poisson.(l.invlink.(fs)))
+(l::PoissonLikelihood)(fs::AbstractVector{<:Real}) = Product(l.(fs))
```
devmotion, line 21:
```suggestion
(l::PoissonLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
```

theogf (commented):
Can't we have a unique ?
```
(l::AbstractLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
```
I think we need 1.6 for this feature but we are already bounding to this version so that seems like a valid thing.

devmotion:
> Can't we have a unique

That's not always correct though so I'm not sure if it's good to define it automatically. Eg, it's incorrect for the `CategoricalLikelihood`.

st--:
I agree it's not going to be valid for _all_ likelihoods, but we could define an AbstractScalarLikelihood (or something like that) for which
```julia
(l::AbstractScalarLikelihood)(fs::AbstractVector{<:Real}) = Product(map(l, fs))
```
would be valid and define it thus?

st--:
> That's not always correct though so I'm not sure if it's good to define it automatically. Eg, it's incorrect for the `CategoricalLikelihood`.

Hm, for `CategoricalLikelihood` it's currently defined the same was for all the others. Is this a bug in the code then ?

src/likelihoods/categorical.jl
```diff
@@ -36,6 +36,4 @@ function (l::CategoricalLikelihood)(f::AbstractVector{<:Real})
     return Categorical(l.invlink(f))
 end
 
-function (l::CategoricalLikelihood)(fs::AbstractVector)
-    return Product(Categorical.(l.invlink.(fs)))
-end
```
st--, line 41: is this a bug then @devmotion ?

st--: looks like this is assuming vectors of vectors (or vectors of tuples)?

st--: Could the following work then:
```julia
(l::AbstractLikelihood)(fs::AbstractVector) = Product(map(l, fs))
(l::CategoricalLikelihood)(f::AbstractVector{<:Real}) = Categorical(l.invlink(f))
```
?

devmotion: No, I don't think this is a bug. It is just the usual dispatch we also use in KernelFunctions - multiple samples form a vector, e.g., `ColVecs`, `RowVecs` or a vector of vectors. The single sample case is handled by the definition for `AbstractVector{<:Real}` above.

willtebbutt (approved):
I'm happy that this is an improvement over the existing implementations, so I'm happy for it to be merged.

@st-- I'm also happy for this to be rolled in with the 0.4 release associated with #70 , but please only merge this when you're about to merge that PR also, so that we don't leave master hanging with unreleased changes.

PR #81, opened 2022-04-19
Implement Analytic Likelihood for Student's T Distribution

WawerOS:


src/likelihoods/studentt.jl
```diff
@@ -0,0 +1,29 @@
+
+using SpecialFunctions :: logbeta
+using IrrationalConstants :: logπ, 
+
+
+"""
+    StudentTLikelihood(σ²,ν)
+
+Student's T likelihood with `σ²` scale and ν degrees of freedom . This is to be used if we assume that the
```
theogf, line 9:
```suggestion
Student's T likelihood with scale `σ²` and `ν` degrees of freedom. This is to be used if we assume that the
```

src/likelihoods/studentt.jl
```diff
@@ -0,0 +1,29 @@
...
+uncertainity associated with the data follows a Student's T distribution.
+
+```math
+    p(y|f) = \\operatorname{Student}(y | f, σ², ν)
+```
+"""
+
+struct StudentTLikelihood{T<:Real, Tn :: Real} <: AbstractLikelihood
+    σ²::Vector{T}
+    ν::Vector{Tn}
```
theogf, line 19: We tend to change the implementation to avoid storing everything in Vector (I know it's not true for `GaussianLikelihood` but it should change).
```suggestion
    σ²::T
    ν::Tn
```

src/likelihoods/studentt.jl
```diff
@@ -0,0 +1,29 @@
...
+end
+
+function expected_loglikelihood( ::AnalyticExpectation,lik::StudentTLikelihood,q_f :: AbstractVector{<:Normal}, y :: AbstractVector{<:Real})
+    f_μ = mean.(q_f)
+    # Why?
+    return sum(-logbeta(0.5,0.5*lik.ν) .- 0.5*logπ .- 0.5*log(lik.ν) .- log(lik.σ²) .- (0.5*(lik.ν+1))*log.(1 .+ ((y .- f_μ).^2 + var.(q_f)) / lik.σ²))
```
theogf, line 25: I am not sure that this is correct. The expectation of `log( (y - f)^2) is not available analytically I think

theogf (commented):
Thanks a lot for the PR!
A few things are missing.
Most importantly, you need to implement 
```
(lik::StudentTLikelihood)(f::Real)
```
also you need to add the file to the list of included files in `src/GPLikelihoods.jl` and add a test in the test suite.

PR #80, opened 2022-04-17
Implement multiple parametrization for Negative Binomial

theogf:
As discussed in #69, we need to be more flexible with the parametrization of `NegativeBinomialLikelihood`.
This is done here by having an extra parameter.
The interface is easily extended by defining a new type and two functions!

Also add docs to explain all this mess

There is still an issue with `Functors.jl` I will try to solve. But I thought I would already show the current status to get some feedback.

PR #82, opened 2022-04-25
negative binomial parametrizations as their own structs, not using a container type

st--:
while discussing #80 
(draft, not to be merged)

### DIFF
--- a/src/likelihoods/negativebinomial.jl
+++ b/src/likelihoods/negativebinomial.jl
@@ -1,13 +1,17 @@
-abstract type NBParam end
+"""
+    NegativeBinomialLikelihood
+
+Abstract base type for the Negative Binomial likelihood. There are multiple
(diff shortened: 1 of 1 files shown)


src/likelihoods/negativebinomial.jl
```diff
@@ -79,12 +63,13 @@ This corresponds to the definition used by [Distributions.jl](https://juliastats
   p(k|successes, f) = \\frac{\\Gamma(k+successes)}{k! \\Gamma(successes)} l(f)^successes (1 - l(f))^k
 ```
 """
-struct NBParamSuccess{T} <: NBParam
+struct NBParamSuccess{T,Tl} <: NegativeBinomialLikelihood
```
st--, line 66: here we could now (well not NOW now, in the future) specialize on "LinkFor[0,1]"

PR #90, opened 2022-06-22
Some performance optimizations

### LINKED ISSUE #77: Question on implementation and type stability
I was studying the implementation of the expected likelihood and stumbled upon this:

https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/blob/498af318aecb891c1497e1a4a19c26dc7ebd0c47/src/expectations.jl#L84-L96

This uses `Broadcast.broadcasted`, which is not documented. If I'm not mistaken, this means that it is not part of Julia's public API, and could break at any point in time. Besides that, I wonder what is the advantage over this:

```julia
function expected_loglikelihood(
    gh::GaussHermiteExpectation, lik, q_f::AbstractVector{<:Normal}, y::AbstractVector
)
    # Compute the expectation via Gauss-Hermite quadrature
    # using a reparameterisation by change of variable
    # (see e.g. en.wikipedia.org/wiki/Gauss%E2%80%93Hermite_quadrature)
    return sum(expected_loglikelihood(gh, lik, q_fᵢ, yᵢ) for (q_fᵢ, yᵢ) in zip(q_f, y))
end
```
This also seems to be type-stable, while the current implementation isn't.

simsurace:
I found this type-stable but allocating version of the Gauss-Hermite `expected_loglikelihood` to be better for AD, compared to the original version and other type-stable implementations I tried, e.g. a hand-written `pairwise_sum(f, itr1, itr2)` (necessary because the multi-argument `mapreduce` does not use pairwise summation, see #77).
I also simplified the parameter conversions in `NegativeBinomialLikelihood(NBParamII(...))`.

This PR includes new tests that check the type stability of `expected_loglikelihood` with all integration methods and increases the number of tested likelihoods.

Closes #77

### DIFF
--- a/src/expectations.jl
+++ b/src/expectations.jl
@@ -87,7 +87,7 @@ function expected_loglikelihood(
     # Compute the expectation via Gauss-Hermite quadrature
     # using a reparameterisation by change of variable
     # (see e.g. en.wikipedia.org/wiki/Gauss%E2%80%93Hermite_quadrature)
-    # PR #86 introduces eager instead of lazy broadcast over observations
+    # PR #90 introduces eager instead of lazy broadcast over observations
     # and Gauss-Hermit points and weights in order to make the function
     # type stable. Compared to other type stable implementations, e.g.
     # using a custom two-argument pairwise sum, this is faster to
--- a/src/expectations.jl
+++ b/src/expectations.jl
@@ -81,31 +81,19 @@ function expected_loglikelihood(
 end
 
 # Compute the expected_loglikelihood over a collection of observations and marginal distributions
-function expected_loglikelihood(
+function GPLikelihoods.expected_loglikelihood(
     gh::GaussHermiteExpectation, lik, q_f::AbstractVector{<:Normal}, y::AbstractVector
 )
     # Compute the expectation via Gauss-Hermite quadrature
     # using a reparameterisation by change of variable
     # (see e.g. en.wikipedia.org/wiki/Gauss%E2%80%93Hermite_quadrature)
-    return sum(Broadcast.instantiate(
-        Broadcast.broadcasted(y, q_f) do yᵢ, q_fᵢ  # Loop over every pair
-            # of marginal distribution q(fᵢ) and observation yᵢ
-            expected_loglikelihood(gh, lik, q_fᵢ, yᵢ)
-        end,
-    ))
-end
-
-# Compute the expected_loglikelihood for one observation and a marginal distributions
-function expected_loglikelihood(gh::GaussHermiteExpectation, lik, q_f::Normal, y)
-    μ = mean(q_f)
-    σ̃ = sqrt2 * std(q_f)
-    return invsqrtπ * sum(Broadcast.instantiate(
-        Broadcast.broadcasted(gh.xs, gh.ws) do x, w # Loop over every
-            # pair of Gauss-Hermite point x with weight w
-            f = σ̃ * x + μ
-            loglikelihood(lik(f), y) * w
-        end,
-    ))
+    # PR #86 introduces eager instead of lazy broadcast over observations
+    # and Gauss-Hermit points and weights in order to make the function
+    # type stable. Compared to other type stable implementations, e.g.
+    # using a custom two-argument pairwise sum, this is faster to
+    # differentiate using Zygote.
+    A = loglikelihood.(lik.(sqrt2 .* std.(q_f) .* gh.xs' .+ mean.(q_f)), y) .* gh.ws'
+    return invsqrtπ * sum(A)
 end
 
 function expected_loglikelihood(
--- a/src/likelihoods/negativebinomial.jl
+++ b/src/likelihoods/negativebinomial.jl
@@ -148,8 +148,9 @@ end
 
 function (l::NegativeBinomialLikelihood{<:NBParamII})(f::Real)
     # Simplify parameter conversions and avoid splatting
-    r = inv(l.params.α)
-    p = inv(one(l.params.α) + l.params.α * l.invlink(f))
+    α = l.params.α
+    r = inv(α)
+    p = inv(one(α) + α * l.invlink(f))
     return NegativeBinomial(r, p)
 end
 
--- a/src/likelihoods/negativebinomial.jl
+++ b/src/likelihoods/negativebinomial.jl
@@ -147,9 +147,10 @@ struct NBParamII{T} <: NBParamMean
 end
 
 function (l::NegativeBinomialLikelihood{<:NBParamII})(f::Real)
-    μ = l.invlink(f)
-    ev = l.params.α * μ
-    return NegativeBinomial(_nb_mean_excessvar_to_r_p(μ, ev)...)
+    # Simplify parameter conversions and avoid splatting
+    r = inv(l.params.α)
+    p = inv(one(l.params.α) + l.params.α * l.invlink(f))
+    return NegativeBinomial(r, p)
 end
 
 """
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "GPLikelihoods"
 uuid = "6031954c-0455-49d7-b3b9-3e1c99afaf40"
 authors = ["JuliaGaussianProcesses Team"]
-version = "0.4.3"
+version = "0.4.4"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/test/expectations.jl
+++ b/test/expectations.jl
@@ -1,6 +1,4 @@
 @testset "expectations" begin
-    # Test that the various methods of computing expectations return the same
-    # result.
     rng = MersenneTwister(123456)
     q_f = Normal.(zeros(10), ones(10))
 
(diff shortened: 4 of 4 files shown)


src/expectations.jl
```diff
@@ -81,31 +81,19 @@ function expected_loglikelihood(
 end
 
 # Compute the expected_loglikelihood over a collection of observations and marginal distributions
-function expected_loglikelihood(
+function GPLikelihoods.expected_loglikelihood(
```
simsurace, line 84:
```suggestion
function expected_loglikelihood(
```

src/likelihoods/negativebinomial.jl
```diff
@@ -147,9 +147,10 @@ struct NBParamII{T} <: NBParamMean
 end
 
 function (l::NegativeBinomialLikelihood{<:NBParamII})(f::Real)
-    μ = l.invlink(f)
-    ev = l.params.α * μ
-    return NegativeBinomial(_nb_mean_excessvar_to_r_p(μ, ev)...)
+    # Simplify parameter conversions and avoid splatting
```
devmotion, line 150: Usually good to avoid splatting 👍 (even though it's a simple case here)

src/likelihoods/negativebinomial.jl
```diff
@@ -147,9 +147,10 @@ struct NBParamII{T} <: NBParamMean
...
+    r = inv(l.params.α)
+    p = inv(one(l.params.α) + l.params.α * l.invlink(f))
```
devmotion, line 152: One could unpack alpha once to reduce the number of field accesses (even though the compiler can optimize it often, I guess).

simsurace:
```suggestion
    α = l.params.α
    r = inv(α)
    p = inv(one(α) + α * l.invlink(f))
```

src/expectations.jl
```diff
@@ -87,25 +87,13 @@ function expected_loglikelihood(
...
-            # pair of Gauss-Hermite point x with weight w
-            f = σ̃ * x + μ
-            loglikelihood(lik(f), y) * w
-        end,
-    ))
+    # PR #86 introduces eager instead of lazy broadcast over observations
+    # and Gauss-Hermit points and weights in order to make the function
+    # type stable. Compared to other type stable implementations, e.g.
+    # using a custom two-argument pairwise sum, this is faster to
+    # differentiate using Zygote.
```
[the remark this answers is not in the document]
simsurace: I think the main part of the optimization is happening in https://github.com/JuliaStats/Distributions.jl/pull/1568, but it cannot be realized unless this function is type-stable. So for now I would go with this type-stable implementation. Funnily enough, it did not affect the allocations of the forward pass, only made the computation time slightly worse (balanced by the rather large speedup of the reverse pass by Zygote). So maybe the compiler does optimize the array allocation away. I'm not sure. In any case, I did not find any drawback with this modification.

I'm hoping that once Enzyme becomes the norm we won't see these performance-sensitivities as often as we do now.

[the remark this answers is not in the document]
devmotion: > So maybe the compiler does optimize the array allocation away. 

I'm pretty sure that doesn't happen, I haven't encountered any examples where the compiler has elided array allocations.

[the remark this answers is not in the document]
simsurace: So maybe they are just insignificant in the grand scheme. I will try on a larger example to see whether they get significant there.

simsurace:
Some benchmarks. This is the code to reproduce the example:

```julia
using BenchmarkTools
using Distributions
using GPLikelihoods
using Zygote

function testfun_small(θ)
    lik = NegativeBinomialLikelihood(NBParamII(θ[end]))
    q_f = Normal.(θ[1:3], θ[4:6])
    y = rand(0:10, 3)
    gh = GPLikelihoods.GaussHermiteExpectation(100)
    return expected_loglikelihood(gh, lik, q_f, y)
end
θ_small = rand(7)

function testfun_large(θ)
    lik = NegativeBinomialLikelihood(NBParamII(θ[end]))
    q_f = Normal.(θ[1:500], θ[501:1000])
    y = rand(0:10, 500)
    gh = GPLikelihoods.GaussHermiteExpectation(100)
    return expected_loglikelihood(gh, lik, q_f, y)
end
θ_large = rand(1001)

@btime testfun_small($θ_small)
@btime Zygote.gradient(testfun_small, $θ_small)

@btime testfun_large($θ_large)
@btime Zygote.gradient(testfun_large, $θ_large)
```
Here are the results for the small example, forward pass / reverse pass:
| Distributions         | GPLikelihoods#master     | [GPLikelihoods#ss/type_stability](https://github.com/simsurace/GPLikelihoods.jl/tree/ss/type_stability) | This PR |
|--------------|-----------|------------|--------------------|
| v0.25.62        | 120μs / 4698μs<br />(87: 35K / 61179: 2.3M)   | 119μs / 5569μs<br />(87: 35K / 13068: 1.2M)    | 122μs / 4345μs<br />(88: 37K / 48415: 1.8M) |
| [#1568](https://github.com/JuliaStats/Distributions.jl/pull/1568 )    | 120μs / 943μs<br />(91: 35K / 73409: 3.1M)     | 118μs / 249μs<br />(87: 38K / 823: 252K)      | 124μs / 179μs<br />(88: 37K / 396: 169K)  |

And here for the large example, forward pass / reverse pass:
| Distributions         | GPLikelihoods#master     | [GPLikelihoods#ss/type_stability](https://github.com/simsurace/GPLikelihoods.jl/tree/ss/type_stability) | This PR |
|--------------|-----------|------------|--------------------|
| v0.25.62        | 3.883ms / 1.203s<br />(91: 55K / 12182325: 507M)   | 3.965ms / 984ms<br />(87: 55K / 10150949: 387M)     | 4.422ms / 830ms<br />(89: 445K / 8004418: 289M) |
| [#1568](https://github.com/JuliaStats/Distributions.jl/pull/1568 )      | 4.372ms / 143ms<br />(90: 55K / 2125823: 184M)      | 4.422ms / 830ms<br />(87: 55K / 91450: 50M)      | 4.366ms / 8.880ms<br />(89: 445K / 4397: 19M)    |

Thus in the larger example the allocations become more significant for the forward pass. But it is still a clear win for the reverse pass.

devmotion:
> Thus in the larger example the allocations become more significant for the forward pass. But it is still a clear win for the reverse pass.

Isn't that an argument for defining an `rrule`?

simsurace:
I think the rrule that was clearly missing is the one from https://github.com/JuliaStats/Distributions.jl/pull/1568. That alone already moves the reverse pass into a more reasonable range. The question is why that is not sufficient. My working hypothesis was that it was due to the type-instability, so I pushed in that direction. I have now sunk a lot of time into this, but I think it has paid off. AD speed is now very well within the reasonable range. This is a complete game changer for fast training of GPs with count observations. The additional allocations in the forward pass don't matter that much for that application (I think in the data size where they would, you would anyway use sparse approximations).

I don't really have fundamental arguments against an even better/simpler solution, but I don't really much time to invest in this right now, given that I already spent that much. What function would you write the `rrule` for? The method for `expected_loglikelihood` that's being called here just contains a `Broadcast.broadcasted`. Would that need an `rrule`? It's unclear to be how one would write such a rule.

Unless a very trivial solution comes to mind, perhaps we can merge this and if the forward pass allocations become an issue later, we can always revisit this.

devmotion:
> What function would you write the `rrule` for? The method for `expected_loglikelihood` that's being called here just contains a `Broadcast.broadcasted`. Would that need an `rrule`? It's unclear to be how one would write such a rule.

For `expected_loglikelihood`, similar to e.g. the rrule for `sum(f, x)` (https://github.com/JuliaDiff/ChainRules.jl/blob/7904019b89f9440925dc8a88ad77e483ebe81619/src/rulesets/Base/mapreduce.jl#L54-L107, without the special optimizations initially I assume). One could just collect the primal values and pullbacks of the individual `expected_loglikelihood` calls, as in https://github.com/JuliaDiff/ChainRules.jl/blob/7904019b89f9440925dc8a88ad77e483ebe81619/src/rulesets/Base/mapreduce.jl#L80-L81. I.e., one would also allocate - but only if one wants to differentiate. And in the pullback one collects and sums everything: https://github.com/JuliaDiff/ChainRules.jl/blob/7904019b89f9440925dc8a88ad77e483ebe81619/src/rulesets/Base/mapreduce.jl#L95-L103.

simsurace:
Thanks for the explanations, but I think that this is currently beyond my abilities/time. Do you consider this to be a blocker?

simsurace:
Are there any blockers for this? It would be great to get this merged, as together with https://github.com/JuliaStats/Distributions.jl/pull/1568 (or its follow-up) this is a real game-changer for performance.

src/expectations.jl
```diff
@@ -87,25 +87,13 @@ function expected_loglikelihood(
...
+    A = loglikelihood.(lik.(sqrt2 .* std.(q_f) .* gh.xs' .+ mean.(q_f)), y) .* gh.ws'
+    return invsqrtπ * sum(A)
 end
```
theogf, line 97: Any reason why you deleted this function?

theogf: Oh it does not render well. It's supposed to point at the one observation case.

theogf: Right, I forgot about the nesting and since none of the other expectation method support it, it does not make sense to keep it.

devmotion (approved):
I still think this is a hack and there should be a better solution (such as fixing Zygote or, even safer, defining an `rrule`). The latter seems preferable since it is not guaranteed that these workarounds here work similarly well for future compiler and Zygote versions.

However, I don't want to block a short-term fix so I'm fine with the PR. I suggest opening an issue so the workaround is not forgotten and more likely to be improved at some point.

simsurace:
Yes I agree it is a hack. I'll open an issue

PR #97, opened 2023-02-23
update compat helper

willtebbutt:
Upgrades compat helper to the latest version of the GHA

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,27 +1,46 @@
-name: CompatHelper
 
+name: CompatHelper
 on:
   schedule:
     - cron: 0 0 * * *
   workflow_dispatch:
-
+permissions:
+  contents: write
+  pull-requests: write
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
     steps:
+      - name: Check if Julia is already available in the PATH
+        id: julia_in_path
+        run: which julia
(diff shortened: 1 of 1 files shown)


.github/workflows/CompatHelper.yml
```diff
@@ -1,27 +1,46 @@
...
           uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
-          version = "2"
+          version = "3"
           Pkg.add(; name, uuid, version)
         shell: julia --color=yes {0}
       - name: "Run CompatHelper"
         run: |
           import CompatHelper
-          CompatHelper.main(; subdirs=["", "test", "docs"])
+          CompatHelper.main()
```
devmotion, line 41: Is this intended?
```suggestion
          CompatHelper.main(; subdirs=["", "test", "docs"])
```

devmotion (approved):
Looks good to me, we should definitely use version 3. I started a CompatHelper run with the changes in this PR to see if everything works correctly: https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/actions/runs/4252591667/jobs/7396376500

Feel free to merge if it passes and my comment below is addressed :slightly_smiling_face:

.github/workflows/CompatHelper.yml
willtebbutt: No. This was not intentional. Good catch

PR #105, opened 2024-03-10
Allow non-constant likelihoods

simsurace:
This PR generalizes all `expected_loglikelihood` methods to also accept `AbstractVector`s, such that it can be used for input-dependent (e.g. heteroskedastic) likelihoods. See e.g. https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/216

### DIFF
--- a/src/expectations.jl
+++ b/src/expectations.jl
@@ -31,29 +31,18 @@ default_expectation_method(_) = GaussHermiteExpectation(20)
         y::AbstractVector,
     )
 
-This function computes the expected log likelihood:
-
+This function computes the sum of the expected log likelihoods:
 ```math
-    ∫ q(f) log p(y | f) df
+    ∑ᵢ ∫ qᵢ(f) log pᵢ(yᵢ | f) df
 ```
-where `p(y | f)` is the process likelihood. This is described by `lik`, which should be a
-callable that takes `f` as input and returns a Distribution over `y` that supports
-`loglikelihood(lik(f), y)`.
-
-`q(f)` is an approximation to the latent function values `f` given by:
-```math
-    q(f) = ∫ p(f | u) q(u) du
-```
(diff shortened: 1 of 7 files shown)


src/expectations.jl
```diff
@@ -31,29 +31,18 @@ default_expectation_method(_) = GaussHermiteExpectation(20)
...
-`q(f)` is an approximation to the latent function values `f` given by:
+This function computes the sum of the expected log likelihoods:
 ```math
-    q(f) = ∫ p(f | u) q(u) du
+    ∑ᵢ ∫ qᵢ(f) log pᵢ(yᵢ | f) df
 ```
-where `q(u)` is the variational distribution over inducing points.
-The marginal distributions of `q(f)` are given by `q_f`.
+where `pᵢ(yᵢ | f)` is the process likelihood and `yᵢ` the observation at location/site `i`.
+The argument `q_f` is the vector of normal distributions `qᵢ(f)`, and the argument `y` is
```
theogf, line 39: Maybe just mention that `qᵢ(f)` corresponds to the normal distribution for f at location.site `i`?

src/expectations.jl
```diff
@@ -94,14 +87,21 @@ function expected_loglikelihood(
...
+
 function expected_loglikelihood(
     ::AnalyticExpectation, lik, q_f::AbstractVector{<:Normal}, y::AbstractVector
 )
     return error(
         "No analytic solution exists for $(typeof(lik)). Use `DefaultExpectationMethod`, `GaussHermiteExpectation` or `MonteCarloExpectation` instead.",
     )
 end
+
+_maybe_ref(lik) = Ref(lik)
```
theogf, line 106: Instead of `_maybe_ref` can't we just make likelihoods iterable?

theogf (commented):
I had a two year old review pending 😅

I'll do another pass now

